PR #53, opened 2016-11-27
Genius/KYE EasyPen M406XE

al42and:
This pull request adds support for Genius EasyPen M406XE.
It was tested with Krita 2.9.7 and GIMP 2.8.16 on Ubuntu 16.04 with 4.8.6-040806-generic kernel and XOrg 1.18.4.

Known problems: 
1. Similarly to Genius EasyPen M610X, the "eraser" button is mapped to "redo".
2. For some reason, with 8581c94 commit (i.e. handling of 0x12 ReportID), Krita stops detecting tablet (and treats it as mouse, i.e. no pressure sensitivity), although according to `xinput` the only thing that changes is additional `XIKeyClass` class. Seems like a bug in Krita/QT. If desirable, I can make it configurable through module parameters.


hid-kye.c
```diff
@@ -483,6 +483,91 @@ static __u8 pensketch_m912_rdesc_fixed[] = {
...
+	0x09, 0x31,         /*          Usage (Y),                  */
+	0x26, 0x00, 0x28,   /*          Logical Maximum (10240),    */
+	0x46, 0x00, 0x28,   /*          Physical Maximum (10240),   */
+	0x81, 0x02,         /*          Input (Variable),           */
+	0x05, 0x0D,         /*          Usage Page (Digitizer),     */
+	0x09, 0x30,         /*          Usage (Tip Pressure),       */
+	0x14,               /*          Logical Minimum (0),        */
+	0x26, 0xFF, 0x03,   /*          Logical Maximum (1023),     */
+	0x34,               /*          Physical Minimum (0),       */
+	0x46, 0xFF, 0x03,   /*          Physical Maximum (1023),    */
```
spbnick: Having Physical Maximum without units doesn't make sense. As nobody actually cares about pressure units, they're not consistent, and everyone just uses pressure curves, it's better to just remove these.

hid-kye.c
```diff
@@ -483,6 +483,91 @@ static __u8 pensketch_m912_rdesc_fixed[] = {
...
+	0x85, 0x12,         /*      Report ID (18),                 */
+	0x14,               /*      Logical Minimum (0),            */
+	0x25, 0x01,         /*      Logical Maximum (1),            */
+	0x75, 0x01,         /*      Report Size (1),                */
+	0x95, 0x04,         /*      Report Count (4),               */
+	0x0A, 0x79, 0x02,   /*      Usage (AC Redo Or Repeat),      */
+	0x0A, 0x1A, 0x02,   /*      Usage (AC Undo),                */
+	0x0A, 0x2D, 0x02,   /*      Usage (AC Zoom In),             */
+	0x0A, 0x2E, 0x02,   /*      Usage (AC Zoom Out),            */
+	0x81, 0x02,         /*      Input (Variable),               */
```
spbnick: You don't need to change Report Size past this point. Just leave it at 1, as set above, and specify the appropriate report count to ignore the rest of the report.

hid-kye.c
```diff
@@ -483,6 +483,91 @@ static __u8 pensketch_m912_rdesc_fixed[] = {
...
+	0x95, 0x01,         /*      Report Count (1),               */
+	0x75, 0x14,         /*      Report Size (4),                */
+	0x81, 0x03,         /*      Input (Constant, Variable),     */
+	0x75, 0x20,         /*      Report Size (32),               */
+	0x81, 0x03,         /*      Input (Constant, Variable),     */
```
spbnick: You don't need to have two inputs either. Just have one Input with Report Count of 36.

hid-kye.c
```diff
@@ -483,6 +483,91 @@ static __u8 pensketch_m912_rdesc_fixed[] = {
...
+	0x25, 0x01,         /*          Logical Maximum (1),        */
+	0x95, 0x01,         /*          Report Count (1),           */
+	0x81, 0x02,         /*          Input (Variable),           */
+	0x05, 0x01,         /*          Usage Page (Desktop),       */
+	0x09, 0x30,         /*          Usage (X),                  */
+	0x75, 0x10,         /*          Report Size (16),           */
+	0x95, 0x01,         /*          Report Count (1),           */
+	0xA4,               /*          Push,                       */
+	0x55, 0x0D,         /*          Unit Exponent (13),         */
+	0x65, 0x33,         /*          Unit (Inch^3),              */
```
spbnick: To verify that it worked, check the resolution in `evtest` output - it should match the specifications (2560 LPI in each direction, I assume).

al42and:
Hi @spbnick, thank you very much for such a thorough review.

I don't have much experience with USB protocol, sorry it was such a mess.

> First, please make it a single commit, this way it will be easier to review.
Sure. I tried to address all your comments, and squashed it all into single commit.

As for Krita issue -- thanks for pointing me towards Xorg logs: for some reason, with buttons enabled the XInput type of the device became `KEYBOARD` instead of `TABLET`, thus confusing Krita/Qt. Breaking the device in two by setting `HID_QUIRK_MULTI_INPUT` resolves this problem, without seemingly breaking anything.
By that logic it seems that this flag should be set for several other tablets as well, but I have no way to test them.

hid-kye.c
al42and: Done. Should I correct similar thingie at line 347 (from Genius M610X descriptor), or should I leave it be, since I can not verify that it works?

hid-kye.c
al42and: Looking at tablet specs, it seems that it reports its physical size in 10^3 cm. If you don't mind, I'd rather leave physical size in metric system (with correct `Unit` field, of course).

hid-kye.c
spbnick: Thanks! Looks much better now! Better leave M610X alone, as I don't have a way to test it either.

hid-kye.c
spbnick: I've never seen a tablet that used metric size or resolution. All of them are in inches. Some shops or OEMs can specify them in mm or cm on their webpages, but the hardware is in round or nice fractions of inches. Manufacturers may also convert resolutions sizes to metric for metric countries.

If you take the logical extents (15360x10240) and divide them by the advertised size (6x4 inches), you'll get exactly the advertised resolution of 2560 LPI.

spbnick:
No problem, that's why I'm here - to help make the code better :)
Thank you for the fixes. Ah, yes, you should have HID_QUIRK_MULTI_INPUT. Most other tablets that need it, already have it set upstream in `drivers/hid/usbhid/hid-quirks.c`.

Except the units, everything seems in order now. Please fix them, oh, and please also change the commit subject to "kye: Add support for EasyPen M406XE".

Thank you.

al42and:
Changed the descriptor to use inches, squashed it all into single commit.

spbnick:
Thanks! This looks fine now. However, I think we lost the `90-hid-rebind.rules` update along the way.

Also, did you test that the resolution is reported appropriately to the userspace?
The evtest output for the pen device should show the correct resolution in lines-per-mm (about 100, I believe).

al42and:
> However, I think we lost the 90-hid-rebind.rules update along the way.

I haven't modified udev rules. The tablet is matched by already existing line that catches all Kye/Genius devices:
`SUBSYSTEM=="hid", ACTION=="add", ENV{HID_ID}=="0003:00000458:*",`

> Also, did you test that the resolution is reported appropriately to the userspace?

Yes, it reports 101 lines/mm = 2565 lines/inch

spbnick:
Ah, please excuse me, I confused your PR with #55.
Everything is fine, then, thank you.

Would you consider adding information on your tablet to the [tablet repo](https://github.com/DIGImend/tablets)? I could then publish it on the DIGImend website, along with support info.

spbnick:
Merged, thank you.

PR #55, opened 2016-12-09
uclogic: add suppor for XP-Pen G540 osu! tablet

slawkis:


spbnick:
Thanks, @slawkis, this looks fine. Only please squash the second commit into the first and add a comment about the vendor ("XP-Pen Technology") to the hid-ids.h file, similarly to the other vendors.

PR #57, opened 2016-12-28
Support for XP-Pen G540.

slawkis:


spbnick:
Thanks a lot, @slawkis! Would you consider adding a directory for your tablet to the [tablets repo](https://github.com/DIGImend/tablets/), so that it would appear in the [tablet list](http://digimend.github.io/tablets/) on the site?

spbnick:
Merged.

PR #85, opened 2017-11-07
Added support for Gaomon M106K tablet.

kolrabi:
I recently got a M106K and the digimend drivers didn't support it, so I added support for it. The pen position, buttons and pressure work just fine. The hardware buttons on the side seem to work as well. No idea if or how I should make the soft buttons work.

I added support for the 0xC8 (high resolution) parameter block which is now preferred for some tablets over the legacy 0x64 one, which the M106K doesn't support. The report has a somewhat different layout and a different id (8 instead of 7).

From what I can tell, the M106K is a rebranded / redesigned Huion tablet and this patch could also make the new 1060 PLUS work, as they seem to have very similar, if not identical, reports.

Please have a look. :)

### DIFF
--- a/hid-uclogic.c
+++ b/hid-uclogic.c
@@ -29,9 +29,10 @@ static inline s32 le24_to_cpu(const __u8* ptr)
 	return ptr[0] | (ptr[1] << 8UL) | (ptr[2] << 16UL);
 }
 
-#define UCLOGIC_RDESC_ID (0x64)
-#define UCLOGIC_HIRES_RDESC_ID (0xC8)
-#define UCLOGIC_HIRES_RDESC_LENGTH (18)
+#define UCLOGIC_PRM_STR_ID (0x64)
+#define UCLOGIC_HIRES_PRM_STR_ID (0xC8)
+#define UCLOGIC_PRM_STR_LENGTH (UCLOGIC_PRM_NUM * sizeof(__le16))
+#define UCLOGIC_HIRES_PRM_STR_LENGTH (18)
 
 /* Size of the original descriptor of WPXXXXU tablets */
 #define WPXXXXU_RDESC_ORIG_SIZE	212
@@ -966,13 +967,12 @@ static int uclogic_input_configured(struct hid_device *hdev,
  * @pbuf:	Location for the kmalloc'ed parameter array with
  * 		UCLOGIC_PRM_NUM elements.
  */
-static int uclogic_enable_tablet(struct hid_device *hdev, __u8 rdescid, __u8 **pbuf)
+static int uclogic_enable_tablet(struct hid_device *hdev, __u8 rdescid, __u8 **pbuf, size_t len)
 {
 	int rc;
 	struct usb_device *usb_dev = hid_to_usb_dev(hdev);
 	struct uclogic_drvdata *drvdata = hid_get_drvdata(hdev);
 	__u8 *buf = NULL;
-	size_t len;
 
 	/*
 	 * Read string descriptor containing tablet parameters. The specific
@@ -980,11 +980,6 @@ static int uclogic_enable_tablet(struct hid_device *hdev, __u8 rdescid, __u8 **p
 	 * driver traffic.
 	 * NOTE: This enables fully-functional tablet mode.
 	 */
-	if (rdescid == UCLOGIC_HIRES_RDESC_ID) {
-		len = UCLOGIC_HIRES_RDESC_LENGTH;
-	} else {
-		len = UCLOGIC_PRM_NUM * sizeof(__le16);
-	}
 	buf = kmalloc(len, GFP_KERNEL);
 	if (buf == NULL) {
 		rc = -ENOMEM;
@@ -1010,23 +1005,26 @@ static int uclogic_enable_tablet(struct hid_device *hdev, __u8 rdescid, __u8 **p
 	}
 
 	drvdata->tablet_enabled = true;
-	*pbuf = buf;
-	buf = NULL;
+	if (pbuf) {
+		*pbuf = buf;
+		buf = NULL;
+	}
 	rc = 0;
 
 cleanup:
+
 	kfree(buf);
 	return rc;
 }
 
-static void uclogic_fill_placeholders(__u8* prdesc, unsigned int rsize, s32* params)
+static void uclogic_fill_placeholders(__u8* prdesc, unsigned int rsize, s32* params, size_t param_count)
 {
 	__u8 *p;
 	s32 v;
 	for (p = prdesc;
 	     p <= prdesc + rsize -4;) {
 		if (p[0] == 0xFE && p[1] == 0xED && p[2] == 0x1D &&
-		    p[3] < UCLOGIC_PH_ID_NUM) {
+		    p[3] < param_count) {
 			v = params[p[3]];
 			put_unaligned(cpu_to_le32(v), (s32 *)p);
 			p += 4;
@@ -1056,11 +1054,10 @@ static int uclogic_probe_tablet(struct hid_device *hdev,
 	s32 resolution;
 
 	/* Enable tablet mode and get raw device parameters */
-	rc = uclogic_enable_tablet(hdev, UCLOGIC_RDESC_ID, &bbuf);
+	rc = uclogic_enable_tablet(hdev, UCLOGIC_PRM_STR_ID, &bbuf, UCLOGIC_PRM_STR_LENGTH);
 	if (rc != 0) {
 		goto cleanup;
 	}
-
 	buf = (__le16*)bbuf;
 
 	/* Extract device parameters */
@@ -1091,7 +1088,7 @@ static int uclogic_probe_tablet(struct hid_device *hdev,
 
 	/* Format fixed report descriptor */
 	memcpy(drvdata->rdesc, rdesc_template_ptr, drvdata->rsize);
-	uclogic_fill_placeholders(drvdata->rdesc, drvdata->rsize, params);
+	uclogic_fill_placeholders(drvdata->rdesc, drvdata->rsize, params, sizeof(params)/sizeof(*params));
 
 	rc = 0;
 
@@ -1111,7 +1108,7 @@ static int uclogic_probe_tablet_hires(struct hid_device *hdev,
 	s32 resolution;
 
 	/* Enable tablet mode and get raw device parameters */
-	rc = uclogic_enable_tablet(hdev, UCLOGIC_HIRES_RDESC_ID, &buf);
+	rc = uclogic_enable_tablet(hdev, UCLOGIC_HIRES_PRM_STR_ID, &buf, UCLOGIC_HIRES_PRM_STR_LENGTH);
 	if (rc != 0) {
 		goto cleanup;
 	}
@@ -1143,8 +1140,7 @@ static int uclogic_probe_tablet_hires(struct hid_device *hdev,
 
 	/* Format fixed report descriptor */
 	memcpy(drvdata->rdesc, rdesc_template_ptr, drvdata->rsize);
-	uclogic_fill_placeholders(drvdata->rdesc, drvdata->rsize, params);
-
+	uclogic_fill_placeholders(drvdata->rdesc, drvdata->rsize, params, sizeof(params)/sizeof(*params));
 	rc = 0;
 
 cleanup:
@@ -1271,7 +1267,7 @@ static int uclogic_probe(struct hid_device *hdev,
 					uclogic_tablet_hires_rdesc_template,
 					sizeof(uclogic_tablet_hires_rdesc_template));
 			drvdata->is_hires = !rc;
-			if (rc) {
+			if (!drvdata->is_hires) {
 				rc = uclogic_probe_tablet(
 						hdev,
 						uclogic_tablet_rdesc_template,
@@ -1388,12 +1384,12 @@ static int uclogic_resume(struct hid_device *hdev)
 
 	/* Re-enable tablet, if needed */
 	if (drvdata->tablet_enabled) {
-		__u8 *buf = NULL;
-		if (drvdata->is_hires)
-			rc = uclogic_enable_tablet(hdev, UCLOGIC_HIRES_RDESC_ID, &buf);
-		else
-			rc = uclogic_enable_tablet(hdev, UCLOGIC_RDESC_ID, &buf);
-		kfree(buf);
+		if (drvdata->is_hires) {
+			rc = uclogic_enable_tablet(hdev, UCLOGIC_HIRES_PRM_STR_ID, NULL, UCLOGIC_HIRES_PRM_STR_LENGTH);
+		}
+		else {
+			rc = uclogic_enable_tablet(hdev, UCLOGIC_PRM_STR_ID, NULL, UCLOGIC_PRM_STR_LENGTH);
+		}
 		if (rc != 0) {
 			return rc;
 		}
--- a/hid-uclogic.c
+++ b/hid-uclogic.c
@@ -24,6 +24,15 @@
 #include "compat.h"
 #include <linux/version.h>
 
+static inline s32 le24_to_cpu(const __u8* ptr)
+{
+	return ptr[0] | (ptr[1] << 8UL) | (ptr[2] << 16UL);
+}
+
+#define UCLOGIC_RDESC_ID (0x64)
+#define UCLOGIC_HIRES_RDESC_ID (0xC8)
+#define UCLOGIC_HIRES_RDESC_LENGTH (18)
+
 /* Size of the original descriptor of WPXXXXU tablets */
 #define WPXXXXU_RDESC_ORIG_SIZE	212
 
@@ -544,6 +553,7 @@ enum uclogic_ph_id {
 /* Report descriptor template placeholder */
 #define UCLOGIC_PH(_ID) UCLOGIC_PH_HEAD, UCLOGIC_PH_ID_##_ID
 #define UCLOGIC_PEN_REPORT_ID	0x07
+#define UCLOGIC_HIRES_PEN_REPORT_ID	0x08
 
 /* Fixed report descriptor template */
 static const __u8 uclogic_tablet_rdesc_template[] = {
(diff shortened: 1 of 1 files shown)


k-jackson:
This adds support for my Huion "New 1060Plus", with a couple small quirks:

1. the 10th button centers the cursor and forgets the cursor's X position
1. the 11th button centers the cursor and forgets the cursor's Y position

I can happily live with it. Thanks a lot for your hard work :+1:

spbnick:
Hi @kolrabi. From a quick glance this looks good! I'll do a proper review on Monday.

hid-uclogic.c
```diff
@@ -948,6 +1019,23 @@ static int uclogic_enable_tablet(struct hid_device *hdev, __le16 **pbuf)
...
 }
 
+static void uclogic_fill_placeholders(__u8* prdesc, unsigned int rsize, s32* params)
+{
+	__u8 *p;
+	s32 v;
+	for (p = prdesc;
+	     p <= prdesc + rsize -4;) {
+		if (p[0] == 0xFE && p[1] == 0xED && p[2] == 0x1D &&
+		    p[3] < UCLOGIC_PH_ID_NUM) {
```
spbnick: If we are passing `params` as the argument, we better pass number of its elements as well, instead of using a global macro.

hid-uclogic.c
```diff
@@ -24,6 +24,15 @@
...
 #include <linux/version.h>
 
+static inline s32 le24_to_cpu(const __u8* ptr)
+{
+	return ptr[0] | (ptr[1] << 8UL) | (ptr[2] << 16UL);
+}
+
+#define UCLOGIC_RDESC_ID (0x64)
+#define UCLOGIC_HIRES_RDESC_ID (0xC8)
+#define UCLOGIC_HIRES_RDESC_LENGTH (18)
```
spbnick: Naming these "RDESC" is confusing. They are not "report descriptors", they are "string descriptors", specifically "string descriptors containing device parameters which are also configuring the device". I would shorten that to something like "PRM_STR".

hid-uclogic.c
```diff
@@ -913,15 +980,19 @@ static int uclogic_enable_tablet(struct hid_device *hdev, __le16 **pbuf)
 	 * driver traffic.
 	 * NOTE: This enables fully-functional tablet mode.
 	 */
-	len = UCLOGIC_PRM_NUM * sizeof(*buf);
+	if (rdescid == UCLOGIC_HIRES_RDESC_ID) {
```
spbnick: This feels hacky and error-prone. I would rather pass the string descriptor buffer length along with the descriptor ID.

hid-uclogic.c
```diff
@@ -1002,17 +1091,59 @@ static int uclogic_probe_tablet(struct hid_device *hdev,
...
-	     p <= drvdata->rdesc + drvdata->rsize - 4;) {
-		if (p[0] == 0xFE && p[1] == 0xED && p[2] == 0x1D &&
-		    p[3] < ARRAY_SIZE(params)) {
-			v = params[p[3]];
-			put_unaligned(cpu_to_le32(v), (s32 *)p);
-			p += 4;
-		} else {
-			p++;
-		}
+	uclogic_fill_placeholders(drvdata->rdesc, drvdata->rsize, params);
```
spbnick: Nice idea to put this in a function.

hid-uclogic.c
```diff
@@ -544,6 +553,7 @@ enum uclogic_ph_id {
 /* Report descriptor template placeholder */
 #define UCLOGIC_PH(_ID) UCLOGIC_PH_HEAD, UCLOGIC_PH_ID_##_ID
 #define UCLOGIC_PEN_REPORT_ID	0x07
+#define UCLOGIC_HIRES_PEN_REPORT_ID	0x08
```
spbnick: I like the "hires" designation - the resolution bump seems to be exactly the reason they changed the string descriptor ID and format.

hid-uclogic.c
```diff
@@ -1135,10 +1266,17 @@ static int uclogic_probe(struct hid_device *hdev,
...
 		if (intf->cur_altsetting->desc.bInterfaceNumber == 0) {
-			rc = uclogic_probe_tablet(
+			rc = uclogic_probe_tablet_hires(
 					hdev,
-					uclogic_tablet_rdesc_template,
-					sizeof(uclogic_tablet_rdesc_template));
+					uclogic_tablet_hires_rdesc_template,
+					sizeof(uclogic_tablet_hires_rdesc_template));
+			drvdata->is_hires = !rc;
+			if (rc) {
```
spbnick: Shall we maybe write this as `if (!drvdata->is_hires)` for clarity?

kolrabi:
I made the changes you suggested, have a look. :)

spbnick:
Thank you @kolrabi! I spent some more time reading and thinking about this. Can we restructure this a bit differently for easier understanding and less code?

E.g. have a function which simply fills a specified parameter array for the given tablet, handling both low-res and hi-res models? It would also accept NULL as parameter array pointer, for use in `uclogic_resume`. We would need however to handle models, which do respond to the "hires" string descriptor request, but respond with garbage, such as a string. Example: https://github.com/DIGImend/tablets/blob/master/Ugee_G5/probe.txt

Perhaps we can set a boolean field in the driver data, and/or pass it as an argument, letting the function know it can try "hires" descriptor. Or have the function try retrieving a normally invalid string descriptor and comparing the returned value to the "hires parameters" one - if it's the same, then it is also invalid, which is the behavior for older tablets, it seems.

Then we can have a single tablet "probe" function, which would call that one and prepare the replacement descriptor according to the retrieved parameters.

What do you think?

spbnick:
Also, do you think you can work with people from #76 and #77 to get their tablets working too? They seem to be using similar initialization and reports, IIRC. Thank you!

kolrabi:
@spbnick Could you explain in more detail what that restructuring would entail? Are we talking about consolidating uclogic_probe_tablet and uclogic_probe_tablet_hires into a single function?

As for the string descriptor returned for Ugee G5, that string ("U608g") looks more like an id to me than garbage. In any case the driver would already reject it and fall back to the low res description because the string descriptor is too short. But we can of course add some more sanity checks.

spbnick:
@kolrabi Kind of like that. Only take out creating the replacement report descriptor from it, declare it something like this:

    static int uclogic_probe_tablet_params(struct hid_device *hdev, s32 *params)

and make it differentiate lowres and hires tablets by itself, or accept another parameter telling which one it is, if that's not possible. Then build the rest around it. Including the resume function, which would pass NULL instead of "params".

Yes, "U608g" looks like an ID, but that's the point. We cannot say for sure if it's possible for a tablet to have an "ID" that's exactly as long as the parameters we expect.

mhagger:
I started reviewing this PR, but noticed that there were several changes lumped together in the first commit, plus there were several formatting problems. So to make it easier to review (and submit upstream), I

* Rebased the branch onto current master
* Split the commits up into smaller steps
* Fixed some formatting problems

and I pushed the results to [my fork](https://github.com/mhagger/digimend-kernel-drivers) as branch `kolrabi-gaomon-m106k`. I hope I've left the functionality unchanged.

I left @kolrabi as the author, as it's still essentially their work. Feel free to steal my branch as the basis for subsequent work if it saves you effort.

I didn't actually review the last (big) commit yet, but hopefully it's easier now that it isn't commingled with other refactoring changes.

mhagger:
This branch (I tested my version of the branch) makes my Huion New 1060 Plus mostly work:

* Position and pressure seem to work fine.
* The tip switch generates button 1 events
* The pen button closer to the tip generates button 2 events, at least when the tip switch is not engaged
* The pen button further from the tip generates button 3 events, at least when the tip switch is not engaged

The twelve tablet hardware buttons generate the following events:

| Row | Side | Events |
| ----- | ----- | ----- |
| top | left | button 1 |
| top | right | button 2 |
| 2 | left | button 3 |
| 2 | right | button 8 |
| 3 | left | button 9 |
| 3 | right | button 10 |
| 4 | left | button 11 |
| 4 | right | button 12 |
| 5 | left | button 13 |
| 5 | right | NO EVENTS |
| bottom | left | NO EVENTS |
| bottom | right | NO EVENTS |

I think that [my hacky experimental changes](https://github.com/DIGImend/digimend-kernel-drivers/issues/77#issuecomment-348668175), which replaces the button report descriptor too, does a better job of handling all of the tablet buttons. I'll see if I can marry that functionality with this PR branch.

It seems that the `wacom` driver refuses to manage this device:

> HID 256c:006e Pad: tablet unknown to libwacom

If I understand correctly, that's because that driver has a whitelist of tablets that it is willing to manage (?). So as far as I can tell, the tests above were run using the `evdev` driver.

spbnick:
libwacom is just a library for describing tablets, its failure to recognize your tablet should not affect the input functionality, only configuration through the GNOME applet. See http://linuxwacom.sourceforge.net/wiki/index.php/Libwacom

anpaza:
@kolrabi where you took the report format described in uclogic_tablet_hires_rdesc_template[]? Notably, I can't get what the "Usage (In Range)" bit 6 in byte 0 does (0x09, 0x32)? Whatever I do with the tablet, it is always zero.

Also, how you found the format of the block returned by UCLOGIC_HIRES_PRM_STR_ID? Do you know the meaning of the fields other than those declared in uclogic_hires_prm_offset? For example, my Huion 1060 plus returns the following identifier:

13 03 70 c6 00 06 7c 00 ff 1f d8 13 03 0c 10 00 01 00

What are "13 03" at start and "d8 13 03 0c 10 00 01 00" after  pressure range value?

spbnick:
@mhagger, @kolrabi I would prefer you guys cooperated on this pull request instead of creating a new one. Now I get to review two PRs :)

@mhagger, could you please just do a review of @kolrabi PR if you find anything which should be fixed? Then @kolrabi would be able to fix it him/herself.

@kolrabi, do you think you would be able to do the code reorganization I described? If this is too much for you, that's OK too. Perhaps somebody will be able to take over.

kolrabi:
@anpaza I just took uclogic_tablet_rdesc_template, copied it and changed it to match the reports I was getting. No idea what that bit is for. As for that parameter string bytes:

- 13: Should be the length of the data (which is odd, the data you pasted is only 18 bytes)
- 03: Means this data is semantically a string
- d8 13: This is a little endian encoded 16 bit integer 5080, this is the resolution of the tablet in lines/inch
- 03 0c 10 00 01 00: I have no idea. My guess would be either this is some reserved space for future use, but then it would probably be all 00. Maybe the 10 00 and 01 00 describe the layout (16x1) of the soft buttons somehow? Maybe the 03 is the total number of buttons on the stylus? Hard to tell.

@spbnick I'd rather someone else had a look at it, as I wouldn't be able to get to it any time soon, with Christmas coming up and all.

asbaklm:
@mhagger 

From my limited understanding, does the "Graphics Tablet" application (typically used with Wacom devices) in Linux System Settings presumably have no effect with the Huion NEW 1060 Plus?  (8192)

I installed the driver from your kolrabi-gaomon-m106k branch but could only get the buttons on the tablet to respond. When I draw on the surface the green LED's on the tablet show no reaction and nothing happens in Krita or when just trying to move the pointer around.

There must be something misconfigured on my laptop but I tried so many things that I'm not sure what it could be. 

Are there any conf files that require specific settings that you are aware of? I have no idea where things are going wrong.

This is what I did:

- Cloned the mhagger fork, checked out the kolrabi branch
- Compiled the drivers & installed them
- Rebooted
- Tablet buttons are responding but I'm not getting any reaction from the pad

xinput -list shows a Huion Pen & Pad device


Configuration files which may be relevant, or not, are there any specific settings you use for these or any other configs which may be missing?

/usr/share/X11/xorg.conf.d/
10-evdev.conf

/etc/depmod.d/
digimend.conf

anpaza:
Since the patch hasn't been applied yet, here's another patch that can be applied on top of kolrabi's work to fix the tablet buttons. It will replace the HID descriptor for the pad sub-device on "hires" tablets with another one.

[huion-button-support.diff.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/1585013/huion-button-support.diff.txt)

With evtest I'm getting all buttons in their place (BTN_0 to BTN_9 and buttons 10 and 11 are displayed as "?").

Another question how useful this would be. I'm getting button 0 and 1 work as left and right mouse buttons respectively... which is not useful at all. Maybe it would make sense to offset tablet buttons to buttons 4..16? At least users won't get unexpected results.

Also, does anyone know any way to map these buttons to anything useful in GIMP, for example? I can see GIMP has direct linux input device support, but this is gross, as it requires user rights to access /dev/input/* directly. Also I'm getting KDE5 segfaulting when I'm running "xinput test" on the pad device, but that's a KDE problem.

anpaza:
I found the meaning of 7th bit in byte at offset 1.
This is the proximity bit. When you approach the stylus to the tablet, the first report has the bit unset:

08 00 97 11 85 6e 00 00 00 00 00 00

All following reports have the bit set, e.g.:
08 80 ad 08 f5 65 00 00 00 00 00 00

However, when you move the stylus away from the tablet, there is no event with bit 7=0 generated :(
This is a blatant firmware bug in the tablet. Unfortunately, it has serious consequences.

Due to the absence of working proximity-in / proximity-out events, GIMP always thinks the stylus is hovering over the tablet. As a result, core mouse does not work in drawing area (e.g. you can't draw with the mouse, and, most importantly, you can't use the scrolling wheel to navigate around the image).

I have an old Wacom Bluetooth tablet, and it behaves differently. When I'm hovering the stylus over the tablet, "core pointer" does not work. But when I move the stylus away from the tablet, mouse restores full functionality.

I was thinking about some "emulation" of proximity events. When the pressure is 0 and events don't come from tablet for some time (e.g. a few tenths of second), it is reasonable to assume the stylus was moved away from tablet (try keeping the stylus hovering close to the tablet so that it doesn't generate motion events... that's almost impossible). So the driver could repeat the last event, but with 7th bit in byte 1 set to 0.

Also, HID description for the hires tablet has to be changed to reflect the correct position of the (buggy) proximity bit.
[proximity-bit-fix.diff.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/1609633/proximity-bit-fix.diff.txt)

Another thing to keep an eye on is to avoid setting drvdata->invert_pen_inrange to true for hires tablets. I have a complete patch but it's intermixed with the previuous patch I've posted.

To make it easier to go forward it would help if the current patches would be applied right now. They somehow work and don't break anything, but there are problems with the tablet itself.

anpaza:
For better understanding, here are two short logs from Xinput subsystem of X11, with the Wacom and Huion tablets:

# xinput test -proximity "Wacom Graphire BT Pen stylus"              
proximity in  a[0]=5021 a[1]=7895 a[2]=0 a[3]=0 a[4]=0 a[5]=-900
motion a[0]=5021 a[1]=7895 a[2]=0 a[3]=0 a[4]=0 a[5]=-900
motion a[0]=5018 a[1]=7891 a[2]=0 a[3]=0 a[4]=0 a[5]=-900
motion a[0]=5012 a[1]=7890 a[2]=0 a[3]=0 a[4]=0 a[5]=-900
motion a[0]=5002 a[1]=7885 a[2]=0 a[3]=0 a[4]=0 a[5]=-900
motion a[0]=4982 a[1]=7873 a[2]=0 a[3]=0 a[4]=0 a[5]=-900
proximity out a[0]=4982 a[1]=7873 a[2]=0 a[3]=0 a[4]=0 a[5]=-900

# xinput test -proximity "Huion Tablet HA60-F400 Pen"
motion a[0]=8683 a[1]=27178
motion a[0]=8672 a[1]=27158
motion a[1]=27137
motion a[0]=8658 a[1]=27113
motion a[1]=27091

The absense of proximity events is the reason why GIMP (and, possibly, other programs) misbehaves with this tablet.

mhagger:
> Since the patch hasn't been applied yet, here's another patch that can be applied on top of kolrabi's work to fix the tablet buttons. It will replace the HID descriptor for the pad sub-device on "hires" tablets with another one.

@anpaza: with your patch, the tablet buttons get mapped to a device that is listed under "Virtual core keyboard", not "Virtual core pointer" as I've usually seen. Perhaps that is the reason for the KDE crashes?

hid-uclogic.c
```diff
@@ -592,6 +603,55 @@ static const __u8 uclogic_tablet_rdesc_template[] = {
...
+	0x27, UCLOGIC_PH(Y_LM), /*          Logical Maximum (PLACEHOLDER),  */
+	0x47, UCLOGIC_PH(Y_PM), /*          Physical Maximum (PLACEHOLDER), */
+	0x81, 0x02,             /*          Input (Variable),               */
+	0xB4,                   /*          Pop,                            */
+	0x09, 0x30,             /*          Usage (Tip Pressure),           */
+	0x27,
+	UCLOGIC_PH(PRESSURE_LM),/*          Logical Maximum (PLACEHOLDER),  */
+	0x81, 0x02,             /*          Input (Variable),               */
+	0x95, 0x02,             /*          Report Count (2),               */
+	0x81, 0x02,             /*          Input (Variable),               */
```
mhagger: This is six bytes of padding, right? Shouldn't it be `Input(Constant, Variable)`?

hid-uclogic.c
```diff
@@ -24,6 +24,16 @@
 #include "compat.h"
 #include <linux/version.h>
 
+static inline s32 le24_to_cpu(const __u8* ptr)
+{
+	return ptr[0] | (ptr[1] << 8UL) | (ptr[2] << 16UL);
+}
+
+#define UCLOGIC_PRM_STR_ID (0x64)
```
mhagger: The parentheses are unnecessary here and also in lines 33  and 35.

hid-uclogic.c
```diff
@@ -939,15 +1005,35 @@ static int uclogic_enable_tablet(struct hid_device *hdev, __le16 **pbuf)
 	}
 
 	drvdata->tablet_enabled = true;
-	*pbuf = buf;
-	buf = NULL;
+	if (pbuf) {
```
mhagger: Maybe document this new feature that `pbuf` can be set to `NULL` if the caller doesn't want the result.

hid-uclogic.c
```diff
@@ -1250,9 +1384,12 @@ static int uclogic_resume(struct hid_device *hdev)
 
 	/* Re-enable tablet, if needed */
 	if (drvdata->tablet_enabled) {
-		__le16 *buf = NULL;
-		rc = uclogic_enable_tablet(hdev, &buf);
-		kfree(buf);
+		if (drvdata->is_hires) {
```
mhagger: The kernel style for blocks of a single line is (regrettably) to omit the braces:

    if (drvdata->is_hires)
            rc = uclogic_enable_tablet(hdev, UCLOGIC_HIRES_PRM_STR_ID, NULL, UCLOGIC_HIRES_PRM_STR_LENGTH);
    else
            rc = uclogic_enable_tablet(hdev, UCLOGIC_PRM_STR_ID, NULL, UCLOGIC_PRM_STR_LENGTH);

anpaza:
If we look into /var/log/Xorg.log, we can see with the old "Pad" device:

```
[ 39230.294] (**) HID 256c:006e Pad: always reports core events
[ 39230.294] (**) evdev: HID 256c:006e Pad: Device: "/dev/input/event16"
[ 39230.294] (--) evdev: HID 256c:006e Pad: Vendor 0x256c Product 0x6e
[ 39230.294] (--) evdev: HID 256c:006e Pad: Found 13 mouse buttons
[ 39230.294] (--) evdev: HID 256c:006e Pad: Found absolute axes
[ 39230.294] (--) evdev: HID 256c:006e Pad: Found x and y absolute axes
[ 39230.294] (--) evdev: HID 256c:006e Pad: Found absolute tablet.
[ 39230.294] (II) evdev: HID 256c:006e Pad: Configuring as tablet
[ 39230.294] (**) evdev: HID 256c:006e Pad: YAxisMapping: buttons 4 and 5
```

The  reason why it finds absolute axes on the "Pad" device are the lines:

```c
        0x05, 0x01,             /*          Usage Page (Generic Desktop),   */
        0x09, 0x30,             /*          Usage (X),                      */
        0x09, 0x31,             /*          Usage (Y),                      */
        0x95, 0x02,             /*          Report Count (2),               */
        0x81, 0x02,             /*          Input (Data,Var,Abs),           */
```

in uclogic_buttonpad_rdesc[] array. However, there are no absolute X and Y coordinates in the report on the newer hires tablets, so the new report descriptor just doesn't contain them. Then, X11 processes the new device as follows:

```
[ 39573.520] (**) Huion Tablet HA60-F400 Pad: always reports core events
[ 39573.520] (**) evdev: Huion Tablet HA60-F400 Pad: Device: "/dev/input/event16"
[ 39573.572] (--) evdev: Huion Tablet HA60-F400 Pad: Vendor 0x256c Product 0x6e
[ 39573.572] (--) evdev: Huion Tablet HA60-F400 Pad: Found 16 mouse buttons
[ 39573.572] (II) evdev: Huion Tablet HA60-F400 Pad: Forcing relative x/y axes to exist.
[ 39573.572] (II) evdev: Huion Tablet HA60-F400 Pad: Configuring as mouse
[ 39573.572] (**) evdev: Huion Tablet HA60-F400 Pad: YAxisMapping: buttons 4 and 5
```

As you can see, it found only mouse buttons, and no axes at all. But for some reason X11 created x/y relative axes to this device and transformed it into a "complete" mouse. Why it then falls under the "core keyboard" category is still unclean to me (although I can't say this is incorrect... the device contains only buttons, what's it if not a keyboard?). I'll look into xorg-evdev driver how I can make it think it's a "tablet" without adding too much bogus fields; although I don't like it but GIMP also doesn't list "keyboard" input devices in its device list :( There are so many bugs related to extended input devices in X11 software, it's like walking on a mine field.

As to why KDE crashes, this is a bug in any case, as it shouldn't crash in any circumstances. We have a perfectly valid keyboard with 12 buttons, what's wrong with this? Well, since only "xinput test" on the pad device triggers the bug, and that's not a common case, but that still leaves the possibility that (dreaming) one day kde developers will fix that.

hid-uclogic.c
anpaza: 4 bytes of padding (two 16-bit values). And yes, it should be 0x81,0x03 (constant, variable).

anpaza:
I have finished the work on emulating proximity events, now it works very well: A synthetic proximity-out event is generated after 250 ms of inactivity, this is fine as the tablet constantly generates events while the stylus is close to tablet. The code is not specific to this tablet, other UCLogic tablets can use it as well (I suppose the bug is present in other models too), it's enough to set the "emulate_proximity" to true in uclogic_prm structure.

Unfortunately, there's also a bug in xorg-x11-drv-evdev which prevents proximity events from being correctly generated. I've hunted down the bug and will submit a patch to Xorg. With it everything works fine:

'''
# xinput test -proximity "Huion HA60-F400 Tablet Pen"
proximity in  a[0]=9432 a[1]=16417
motion a[0]=9432 a[1]=16417
...
motion a[0]=8683 a[1]=16073
proximity out a[0]=8683 a[1]=16073
'''

This solved the problem with GIMP being unable to use core pointer when tablet is enabled in input devices.

So, what do we do now? I think the original patch can be applied, it kind of works, then I will be able to submit my pull requests so that they can be reviewed and merged as well. Are there any reasons not to do this way?

anpaza:
As to why X11 classifies a "button-only" device as a keyboard, this is also a bug in xorg-evdev driver.
I have submitted a patch for it here: https://bugs.freedesktop.org/show_bug.cgi?id=104630
As I said, extended input devices in linux are a mine field...
Now GIMP lists the pad device under "extended input devices" but anyway you can't do anything useful with it, as GIMP doesn't support assigning functions to extended input device buttons, only to keys. This should be handled with the glib and gimp guys, but I haven't got the guts for it yet.

anpaza:
@kolrabi I had the impudence to re-apply your patches against current git head to solve conflicts, then I added on top some more work for this family of tablets, and submitted as a new pull request #114. I've retained your authorship on your patches, of course.
I hope this doesn't offend you.

kolrabi:
Since my changes are included in pull request #114, I am closing this one. Good work everyone.

PR #90, opened 2017-11-28
uclogic: Add support for Ugee 2150

pmjobin:


spbnick:
Thank you, @pmjobin! Would you care to tell if tablet is fully supported with this commit, or if there's something still not working?

pmjobin:
So far, so good. I'm no expert but I've been able to use it as a pen device for Gimp and no calibration was necessary (other than "xinput map-to-output" because I'm on a dual screen setup). The pressure sensitivity also works well.

Thank you,
PM

spbnick:
Ah, good. Thank you. Would you mind contributing a description of your tablet to the [tablets directory](https://github.com/DIGImend/tablets), so that it could appear on the website, and show your tablet as supported? Thanks!

pmjobin:
Yeah, no problem. I'll do it as soon as i get back home.

Issue #94, opened 2017-12-05
Extract uclogic report descriptors to a separate file


spbnick:
The report descriptors in hid-uclogic.c are taking a lot of lines, extract them to a separate file.

PR #97, opened 2017-12-05
Extracted the ulogic report descriptors

Garethp:
Solved #94

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 obj-m := hid-kye.o hid-uclogic.o hid-polostar.o hid-viewsonic.o
+hid-uclogic-objs := hid-uclogic-rdescs.o
 KVERSION := $(shell uname -r)
 KDIR := /lib/modules/$(KVERSION)/build
 PWD := $(shell pwd)
--- a/hid-uclogic-rdescs.c
+++ b/hid-uclogic-rdescs.c
@@ -1,0 +1,665 @@
+#include <linux/device.h>
+#include "hid-uclogic-rdescs.h"
+
+/* Fixed WP4030U report descriptor */
+const __u8 uclogic_wp4030u_rdesc_fixed[] = {
+        0x05, 0x0D,         /*  Usage Page (Digitizer),             */
+        0x09, 0x02,         /*  Usage (Pen),                        */
+        0xA1, 0x01,         /*  Collection (Application),           */
+        0x85, 0x09,         /*      Report ID (9),                  */
+        0x09, 0x20,         /*      Usage (Stylus),                 */
+        0xA0,               /*      Collection (Physical),          */
+        0x75, 0x01,         /*          Report Size (1),            */
+        0x09, 0x42,         /*          Usage (Tip Switch),         */
+        0x09, 0x44,         /*          Usage (Barrel Switch),      */
+        0x09, 0x46,         /*          Usage (Tablet Pick),        */
+        0x14,               /*          Logical Minimum (0),        */
+        0x25, 0x01,         /*          Logical Maximum (1),        */
+        0x95, 0x03,         /*          Report Count (3),           */
+        0x81, 0x02,         /*          Input (Variable),           */
+        0x95, 0x05,         /*          Report Count (5),           */
+        0x81, 0x01,         /*          Input (Constant),           */
+        0x75, 0x10,         /*          Report Size (16),           */
+        0x95, 0x01,         /*          Report Count (1),           */
+        0x14,               /*          Logical Minimum (0),        */
+        0xA4,               /*          Push,                       */
+        0x05, 0x01,         /*          Usage Page (Desktop),       */
+        0x55, 0xFD,         /*          Unit Exponent (-3),         */
+        0x65, 0x13,         /*          Unit (Inch),                */
+        0x34,               /*          Physical Minimum (0),       */
+        0x09, 0x30,         /*          Usage (X),                  */
+        0x46, 0xA0, 0x0F,   /*          Physical Maximum (4000),    */
+        0x26, 0xFF, 0x7F,   /*          Logical Maximum (32767),    */
+        0x81, 0x02,         /*          Input (Variable),           */
+        0x09, 0x31,         /*          Usage (Y),                  */
+        0x46, 0xB8, 0x0B,   /*          Physical Maximum (3000),    */
+        0x26, 0xFF, 0x7F,   /*          Logical Maximum (32767),    */
+        0x81, 0x02,         /*          Input (Variable),           */
+        0xB4,               /*          Pop,                        */
+        0x09, 0x30,         /*          Usage (Tip Pressure),       */
+        0x26, 0xFF, 0x03,   /*          Logical Maximum (1023),     */
+        0x81, 0x02,         /*          Input (Variable),           */
+        0xC0,               /*      End Collection,                 */
+        0xC0                /*  End Collection                      */
+};
+
+/* Fixed WP5540U report descriptor */
+const __u8 uclogic_wp5540u_rdesc_fixed[] = {
+        0x05, 0x0D,         /*  Usage Page (Digitizer),             */
+        0x09, 0x02,         /*  Usage (Pen),                        */
+        0xA1, 0x01,         /*  Collection (Application),           */
+        0x85, 0x09,         /*      Report ID (9),                  */
+        0x09, 0x20,         /*      Usage (Stylus),                 */
+        0xA0,               /*      Collection (Physical),          */
+        0x75, 0x01,         /*          Report Size (1),            */
+        0x09, 0x42,         /*          Usage (Tip Switch),         */
+        0x09, 0x44,         /*          Usage (Barrel Switch),      */
+        0x09, 0x46,         /*          Usage (Tablet Pick),        */
+        0x14,               /*          Logical Minimum (0),        */
+        0x25, 0x01,         /*          Logical Maximum (1),        */
+        0x95, 0x03,         /*          Report Count (3),           */
+        0x81, 0x02,         /*          Input (Variable),           */
+        0x95, 0x05,         /*          Report Count (5),           */
+        0x81, 0x01,         /*          Input (Constant),           */
+        0x75, 0x10,         /*          Report Size (16),           */
+        0x95, 0x01,         /*          Report Count (1),           */
+        0x14,               /*          Logical Minimum (0),        */
+        0xA4,               /*          Push,                       */
+        0x05, 0x01,         /*          Usage Page (Desktop),       */
+        0x55, 0xFD,         /*          Unit Exponent (-3),         */
+        0x65, 0x13,         /*          Unit (Inch),                */
+        0x34,               /*          Physical Minimum (0),       */
+        0x09, 0x30,         /*          Usage (X),                  */
+        0x46, 0x7C, 0x15,   /*          Physical Maximum (5500),    */
+        0x26, 0xFF, 0x7F,   /*          Logical Maximum (32767),    */
+        0x81, 0x02,         /*          Input (Variable),           */
+        0x09, 0x31,         /*          Usage (Y),                  */
+        0x46, 0xA0, 0x0F,   /*          Physical Maximum (4000),    */
+        0x26, 0xFF, 0x7F,   /*          Logical Maximum (32767),    */
+        0x81, 0x02,         /*          Input (Variable),           */
+        0xB4,               /*          Pop,                        */
+        0x09, 0x30,         /*          Usage (Tip Pressure),       */
+        0x26, 0xFF, 0x03,   /*          Logical Maximum (1023),     */
+        0x81, 0x02,         /*          Input (Variable),           */
+        0xC0,               /*      End Collection,                 */
+        0xC0,               /*  End Collection,                     */
+        0x05, 0x01,         /*  Usage Page (Desktop),               */
+        0x09, 0x02,         /*  Usage (Mouse),                      */
+        0xA1, 0x01,         /*  Collection (Application),           */
+        0x85, 0x08,         /*      Report ID (8),                  */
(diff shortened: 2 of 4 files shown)


Garethp:
Hi @spbnick 

Is this what you had in mind?

Garethp:
For consistency I went ahead and extracted the report descriptors for the other vendors. I know that polostar and viewsonic only have one descriptor each, but I thought that maintaining consistency in where the descriptors are defined would be a good idea. Let me know if you want me to revert them so that this PR only extracts the descriptors for ulogic devices

spbnick:
Thank you, @Garethp for taking on this. This is a good start, but we'll need to do a bunch of things to really get this into shape. First of all, everything we do in the driver should be in view of future contribution back to the kernel, so we should adhere to the kernel standards and expectations as much as possible, or we'll make life difficult for ourselves. I'll understand if you find this frustrating and decide to give it a pass.

Here are some major things which need to be changed:
* I think that separate drivers should use common includes for things which are specific for them. I.e. there shouldn't be a file with just all the descriptors thrown in. Each driver should have their own file with descriptors
* I understand the value of consistency you were going for here, but I think we should only focus on extracting descriptors from drivers which have lots of them. Otherwise when we contribute to the kernel we'll either need to extract report descriptors from *all* the drivers there (and that's a lot), which would produce a lot of files, or told not to extract the small pieces, which is much more likely, and is what I'm asking you to do.
* While macros need to be in headers, the report descriptors themselves should be in .c files. Otherwise each driver including those headers will get their own copy of the descriptors.
* For that matter, we'll need to make kernel modules build from several object files (one main file, one report descriptors), for drivers which have report descriptors extracted. AFAIU, that is done by adding a <module>-objs variable to the Makefile listing additional object files.
* Since the report descriptors will be external symbols, we'll need to prefix them with "uclogic_", "kye_", etc. so they don't clash with other modules or the kernel itself.

Garethp:
Okay, that makes sense. I'll undo a couple of my commits. Do you think the kye descriptors are large enough to warrant their own file? My extracted file for kye descriptors was 538 lines.

Are there any other kernel standards I should be aware of? For example, I'm more used to using longer, more descriptive variable names, but I'm assuming shorter names are part of what's expected for kernel code.

spbnick:
There aren't really very strict guidelines about variable names, but generally they are expected to be terse. Simply take a look at the similar kernel code (in drivers/hid), and read the coding guidelines.

spbnick:
I think hid-kye.c can stay the same, as it consist mostly of report descriptors and there's not much code otherwise.

spbnick:
Ah, good, you got the symbols renamed. However, generally, people don't #include .c files without a good justification, and certainly not in Linux HID drivers. You need to build a separate object file from your .c file, link with it, and have a corresponding .h file.

spbnick:
Look for .h files in Linux kernel sources in `drivers/hid` directory and see how they and corresponding .c files are used. The `wacom_wac.h` and `wacom_wac.c` pair is a good example.

Garethp:
Thanks. I'll take a look and ping you when I've fixed this up

spbnick:
@Garethp, do you think you can finish this before the next weekend? I would like to get on with refactoring and syncing with upstream, as I'd rather do that before adding support for any more tablets to hid-uclogic.c. If you don't have time, I can do this myself, no problem. This is financed, after all.

Garethp:
Sorry for the delays. I've had less time to look at it in the last couple of days than I thought I would have. I'll try to finish it off tomorrow. If I can't, I'll let you complete it.

Garethp:
@spbnick So I *think* I managed to achieve what you were describing. However, when compiling, I get warnings about how I'm defining variables without using them. Is the solution to that to add `extern` declarations in the `.h` file? 

Also, if I totally went in the wrong direction in terms of linking them by `-obj` variables, feel free to just take over and fix it up, I'll just watch what you did to get it how you wanted it

spbnick:
I think you did what needed to be done in the Makefile. I would only put the "-objs" line below the list of modules (the "obj-m" file).

Regarding the modules themselves and the declarations, it's a convention to put variable declarations with `extern` in front of them in the .h file, and the variable definitions (with values) without `static` in the .c file. I.e. remove `static ` from the .c file, and copy variable declarations (without values) from the .c file to the .h file, replacing `static` with `extern`.

We would also need to have a shorter name for the module. How about `hid-uclogic-rdescs.c`?

However, unless you want to learn all this (which I would absolutely welcome and glad to help you with), I can do all that and more myself, maybe this weekend.

Garethp:
@spbnick I'm definitely interested in learning more, so long as I don't slow you down too much doing so. I've moved the `#define` in to the `.h` file (I just realised that those are what you meant by the macros). I've noticed that some of the descriptors are `const` and some are not. Is that right? Shouldn't they all be constants?

spbnick:
I'll be glad to help you along, and so far we're doing well. I'd say all descriptors should be `const`. If that's not the case compilation would fail. I'll take a proper look at your updates in the evening, perhaps.

Garethp:
Well, I'm not sure why it wasn't failing, but I've swapped all definitions over to `const`

spbnick:
Great! I meant it would fail if they couldn't be const.

hid-uclogic-rdescs.h
```diff
@@ -0,0 +1,48 @@
...
+#define UCLOGIC_WPXXXXU_RDESC_ORIG_SIZE	212
+
+/* Size of the original descriptor of the new WP5540U tablet */
+#define UCLOGIC_WP5540U_V2_RDESC_ORIG_SIZE	232
+
+/* Size of the original descriptor of WP1062 tablet */
+#define UCLOGIC_WP1062_RDESC_ORIG_SIZE	254
+
+/* Size of the original descriptor of PF1209 tablet */
+#define UC_LOGIC_PF1209_RDESC_ORIG_SIZE	234
```
spbnick: Could you please write `UCLOGIC` here, without the underscore, to match all the other names?

hid-uclogic-rdescs.h
```diff
@@ -0,0 +1,48 @@
...
+extern const __u8 uclogic_wp1062_rdesc_fixed[];
+extern const __u8 uclogic_pf1209_rdesc_fixed[];
+extern const __u8 uclogic_twhl850_rdesc_fixed0[];
+extern const __u8 uclogic_twhl850_rdesc_fixed1[];
+extern const __u8 uclogic_twhl850_rdesc_fixed2[];
+extern const __u8 uclogic_twha60_rdesc_fixed0[];
+extern const __u8 uclogic_twha60_rdesc_fixed1[];
+extern const __u8 uclogic_tablet_rdesc_template[];
+extern const __u8 uclogic_ugee_ex07_rdesc_template[];
+extern const __u8 uclogic_buttonpad_rdesc[];
```
spbnick: Could you please move the comments describing these here, from the .c file? And then also group them together with the macros, by the tablet model?

Makefile
```diff
@@ -1,3 +1,4 @@
+hid-uclogic-objs := hid-uclogic-rdescs.o
```
spbnick: Could you please move this line below the following one? This way we'll still have the list of modules at the top and their details will follow.

hid-uclogic-rdescs.h
spbnick: I mean group not just the comments, but the variable declarations together with comments.

Garethp:
@spbnick Styling changes applied

hid-uclogic-rdescs.h
```diff
@@ -0,0 +1,76 @@
+#ifndef UCLOGIC_REPORT_DESCRIPTORS
+#define UCLOGIC_REPORT_DESCRIPTORS
```
spbnick: Could you please name this macro after the file, with underscore in front? I.e. `_HID_UCLOGIC_RDESCS_H`?

hid-uclogic-rdescs.h
```diff
@@ -0,0 +1,76 @@
...
+
+#include <linux/device.h>
```
spbnick: Hmm, is there are reason to include this here? Nothing in this file seems to be using it.

hid-uclogic-rdescs.h
```diff
@@ -0,0 +1,76 @@
...
+/* Fixed PF1209 report descriptor */
+extern const __u8 uclogic_pf1209_rdesc_fixed[];
+
+/* Size of the original descriptors of TWHL850 tablet */
+#define UCLOGIC_TWHL850_RDESC_ORIG_SIZE0	182
+
+/* Fixed PID 0522 tablet report descriptor, interface 0 (stylus) */
+extern const __u8 uclogic_twhl850_rdesc_fixed0[];
+
+#define UCLOGIC_TWHL850_RDESC_ORIG_SIZE1	161
```
spbnick: Could you please add a comment here?

hid-uclogic-rdescs.h
```diff
@@ -0,0 +1,76 @@
...
+
+/* Fixed PID 0522 tablet report descriptor, interface 1 (mouse) */
+extern const __u8 uclogic_twhl850_rdesc_fixed1[];
+
+#define UCLOGIC_TWHL850_RDESC_ORIG_SIZE2	92
```
spbnick: And here?

hid-uclogic-rdescs.h
```diff
@@ -0,0 +1,76 @@
...
+/* Fixed PID 0522 tablet report descriptor, interface 2 (frame buttons) */
+extern const __u8 uclogic_twhl850_rdesc_fixed2[];
+
+/* Size of the original descriptors of TWHA60 tablet */
+#define UCLOGIC_TWHA60_RDESC_ORIG_SIZE0 254
+
+/* Fixed TWHA60 report descriptor, interface 0 (stylus) */
+extern const __u8 uclogic_twha60_rdesc_fixed0[];
+
+#define UCLOGIC_TWHA60_RDESC_ORIG_SIZE1 139
```
spbnick: And here, just for consistency?

hid-uclogic-rdescs.h
```diff
@@ -0,0 +1,76 @@
...
+/* Fixed report descriptor template */
+extern const __u8 uclogic_tablet_rdesc_template[];
+
+/* Fixed report descriptor template for Ugee EX07 */
+extern const __u8 uclogic_ugee_ex07_rdesc_template[];
+
+/* Fixed virtual pad report descriptor */
+extern const __u8 uclogic_buttonpad_rdesc[];
+
+#endif
```
spbnick: The convention for long preprocessor conditions is to put the condition in a comment after the `#endif`. I.e.

    #endif /* _HID_UCLOGIC_RDESCS_H */

Could you do that, please?

Garethp:
Sorry, I've gotten a bit sick. I probably won't be able to tackle these until Thursday. Just letting you know, so you don't think I've abandoned this

spbnick:
No problem, thanks for an update, Gareth! We're entering Christmas Time anyway, so everything is going to be on hold.

spbnick:
Get well soon!

Garethp:
@spbnick Comments added and commits squashed

hid-uclogic-rdescs.c
```diff
@@ -0,0 +1,665 @@
...
+        0x09, 0x31,         /*          Usage (Y),                  */
+        0x46, 0xB8, 0x0B,   /*          Physical Maximum (3000),    */
+        0x26, 0xFF, 0x7F,   /*          Logical Maximum (32767),    */
+        0x81, 0x02,         /*          Input (Variable),           */
+        0xB4,               /*          Pop,                        */
+        0x09, 0x30,         /*          Usage (Tip Pressure),       */
+        0x26, 0xFF, 0x03,   /*          Logical Maximum (1023),     */
+        0x81, 0x02,         /*          Input (Variable),           */
+        0xC0,               /*      End Collection,                 */
+        0xC0                /*  End Collection                      */
```
spbnick: The Linux coding guidelines call for using TABs for indentation. Please replace the spaces with TABs in the initial indentation of this and following report descriptors, so that they match the original code in hid-uclogic.c. You can checkout the exact changes between the files with e.g. `vimdiff <(git show origin/master:hid-uclogic.c) hid-uclogic-rdescs.c)` and edit the file to match at the same time.

hid-uclogic-rdescs.c
```diff
@@ -0,0 +1,665 @@
...
+        0x14,               /*          Logical Minimum (0),        */
+        0xA4,               /*          Push,                       */
+        0x05, 0x01,         /*          Usage Page (Desktop),       */
+        0x55, 0xFD,         /*          Unit Exponent (-3),         */
+        0x65, 0x13,         /*          Unit (Inch),                */
+        0x34,               /*          Physical Minimum (0),       */
+        0x09, 0x30,         /*          Usage (X),                  */
+        0x46, 0x10, 0x27,   /*          Physical Maximum (10000),   */
+        0x27, 0x3F, 0x9C,
+        0x00, 0x00, /*          Logical Maximum (39999),    */
```
spbnick: Take note of the change of indentation of this particular line.

hid-uclogic-rdescs.c
```diff
@@ -0,0 +1,665 @@
...
+
+/* Report descriptor template placeholder IDs */
+enum uclogic_ph_id {
+    UCLOGIC_PH_ID_X_LM,
+    UCLOGIC_PH_ID_X_PM,
+    UCLOGIC_PH_ID_Y_LM,
+    UCLOGIC_PH_ID_Y_PM,
+    UCLOGIC_PH_ID_PRESSURE_LM,
+    UCLOGIC_PH_ID_NUM
+};
```
spbnick: The declaration of this enum should appear in `hid-uclogic-rdescs.h` instead (please choose a good spot for it). I'm not even sure why it works as is. Also, please note the indentation.

spbnick:
Superseded by #108.

PR #99, opened 2017-12-10
Small fixes to packaging, tested install on Ubuntu 16.04

doctormo:
I found one minor issue and one housekeeping issue which the patch should fixes.


debian/rules
```diff
@@ -16,6 +16,7 @@ include /usr/share/dpkg/default.mk
 # package maintainers to append LDFLAGS
 #export DEB_LDFLAGS_MAINT_APPEND = -Wl,--as-needed
 
+version := $(shell dpkg-parsechangelog | grep '^Version:' | cut -d' ' -f2 | cut -d- -f1)
```
spbnick: Do Debian guidelines allow using `awk` here? Because then it would be just one command instead of `grep` and two `cut`'s:

    awk -F'[ -]' '/^Version:/ {print $2}'

Or would grep and cut just be easier to understand?

doctormo: I've brought the line in from other packages, I suspect awk would work just fine though.

spbnick: Ah, alright. Could you please change it to `awk`, then, if there's nothing stopping us? Thank you.

doctormo:
Awk didn't work, but sed did. I hope this is ok.

debian/rules
```diff
@@ -16,6 +16,7 @@ include /usr/share/dpkg/default.mk
 # package maintainers to append LDFLAGS
 #export DEB_LDFLAGS_MAINT_APPEND = -Wl,--as-needed
 
+version := $(shell dpkg-parsechangelog --show-field Version | cut -d'-' -f1)
```
spbnick: Hey, this is much better :)!

doctormo:
It's currently in four commits. Do you mean three merge requests?

debian/rules
doctormo: I know right!

spbnick:
I think if you just squash "PPAs require version strings..." into "Fix static version to use dynamic version variable" that will be fine.

spbnick:
And, yeah, within this PR, of course.

doctormo:
Try now.

spbnick:
Thanks a lot, Martin, merged!

PR #108, opened 2018-01-14
uclogic: Extract report descriptors to a module

spbnick:
This is based on and to replace #97, with amendments and additions. @Garethp, could you do a review please? Thank you.


Garethp:
Sorry I didn't get around to working on my PR sooner. Still getting over sickness from December.

It looks good. I suppose it's not an issue, but is there a reason why you moved the `sizeof` calculations to constants? They would execute and calculate the `sizeof` devices that wouldn't be needed, wouldn't it?

Garethp:
Also unimportant, but I noticed that some of the `ORIG_SIZE` definitions have their values column aligned, and others don't. Do you have a preference?

spbnick:
Thanks for the review, Gareth! No problem about the PR.

The sizeof variables are needed because otherwise hid-uclogic.c wouldn't know the sizes of the report descriptors. I'm not sure why it compiled for you. Perhaps some GCC feature is turned off for the kernel, but the code shouldn't actually work, even though it compiles.

The ORIG_SIZE macros are not aligned because I copied them verbatim from hid-uclogic.c. I added another commit aligning them.

Does it look alright to you now?

Garethp:
It all looks good

spbnick:
Merged. Thank you, Gareth!

PR #113, opened 2018-01-19
Joint work by me and Björn Paetzel <kolrabi@kolrabi.de> to support the new hi-resolution Huion tablets.

anpaza:
Andrew Zabolotny <zapparello@ya.ru>:
    added proper support for hi-res tablet pad (12 plain buttons)
    added support for emulating proximity-out events

Björn Paetzel <kolrabi@kolrabi.de>:
    added explicit buffer length to uclogic_enable_tablet,
    added parameter count uclogic_fill_placeholders,
    some readability changes
    Added support for Gaomon M106K tablet.


anpaza:
This pull request includes pull request #85 which is broken now after latest rdesc splitting to separate files.

I've ported the kolrabi's changes to the new head and added my work regarding tablet pad and proximity emulation. Without the later, GIMP 2.8 does not work properly with tablet & mouse.

spbnick:
Impressive work guys! There would need to be some more before I could accept this, though. First of all, are you absolutely sure the tablet doesn't report proximity? Perhaps that could be somehow enabled? Have you take a look at USB traffic with the Windows driver? Nowadays that is very rare and it's best to avoid the complexity you introduce with the emulation. Moreover, at least at some point, the evdev driver managed alright without proximity reports. I have an older tablet that didn't have it and it worked. See if that still works. I'm not sure about the libinput driver.

Next, please don't lump all the changes you did in a single commit. That makes it much harder to figure out what you were changing. Make one commit per logical change as the Linux kernel coding guidelines describe.

Lastly, sorry I rarely answer. It's difficult for me to organize free time for the project.

Thank you.

anpaza:
1. Yes, I'm absolutely sure. Doing 'cat /sys/kernel/debug/hid/.../events' shows that the first event when approaching the pen to tablet is generated with bit 7 in byte 1 being 0, and the rest of the events are with bit 7=1.  But, unfortunately, there's no "closing" event with bit 7=0 when you move the pen away from the tablet. This is a obvious firmware bug. Yes, I examined the windows traffic of the usb driver, it's absolutely the same.

Moreover, if you move the pen quickly away, you won't even get the event with pressure=0, so in such cases software will continue to think that the pen is half-pressed on the tablet, while actually you moved it away.

You can see a confirmation of what I say in your git module "tablets",  files Huion_New_1060_Plus/dumps/{pen_buttons.txt,pen_coords.txt,pen_pressure.txt}.

2. If you won't report proximity (in fact, this means linux event layer has to generate a press/release of "key" BTN_TOOL_PEN), the tablet kind of works, but you can't use the mouse at the same time with the tablet. For example, try to paint something in GIMP with the tablet, then try to paint with the "Core Pointer" mouse (e.g. without enabling extended input device in GIMP's input devices dialog). You can neither paint nor scroll the image with mouse wheel - that's because GIMP "thinks" the pen is still down. After I fixed the proximity events, the mouse works just fine.

3. Okay, I will rollback my patch and apply it in several steps. I'm not a experienced GIT user and I'm not sure I'll be able to retain Björn Paetzel authorship in the commits, I'll mention him in commit messages.

spbnick:
I see, that's really unfortunate. Let's try to keep the proximity emulation changes contained and independent from the rest of the driver so the impact is low. I think it's OK, if you just refer to @kolrabi work, unless he objects. However, please keep changes logically independent, and not just historically accurate. Thank you!

mhagger:
> 2\. I'm not a experienced GIT user and I'm not sure I'll be able to retain Björn Paetzel authorship in the commits

Commit using `git commit --author='Björn Paetzel <kolrabi@kolrabi.de>' [...]`. Of course you should make sure that Björn is OK with the changes that you attribute to him.

The kernel also has a strict requirement that patches be signed off by everybody in their chain of custody. Search for "Signed-off-by" in [this document](https://github.com/torvalds/linux/blob/master/Documentation/process/submitting-patches.rst). I'm not sure whether the digimend project imposes the same requirement, but if so, you would carry over the "Signed-off-by" signature from Björn's original commit followed by your own "Signed-off-by" line stating that you also agree and have the right to pass the patch along.

PR #114, opened 2018-01-23
Joint work by me and Björn Paetzel <kolrabi@kolrabi.de> to support the new hi-resolution Huion tablets

anpaza:
This is a reworked pull request #113.

All incremental changes have been isolated in separate commits. All intermediate repository states are compilable and tested to work.

Also I managed to catch & fix yet another firmware bug - when you move the pen quickly off the tablet, not only pressure is not reported as 0 (leaving tablet-aware applications thinking that pen is still applied with some non-zero pressure), but even the tip switch is not reported as zero, which prevents X11 from generating a button release event, thus even non-tablet-aware apps will be misleaded.

Exercise: with the old driver, open a terminal and mark some text with the pen. Keep it pressed tight, then very quickly move it away from the tablet. The pressure=0 and tip switch=0 are not reported, thus terminal is left with an active mouse grab. Now try to move the regular mouse around (without pressing any buttons), the selection will follow the cursor. Also, you can't exit this mouse grab unless you again put the stylus on the tablet and gently move it away..

The correct event sequence now look this way in evtest:

```
Event: time 1516747806.799359, type 3 (EV_ABS), code 0 (ABS_X), value 1746
Event: time 1516747806.799359, type 3 (EV_ABS), code 1 (ABS_Y), value 12868
Event: time 1516747806.799359, type 3 (EV_ABS), code 24 (ABS_PRESSURE), value 3764
Event: time 1516747806.799359, -------------- SYN_REPORT ------------
Event: time 1516747806.803357, type 3 (EV_ABS), code 0 (ABS_X), value 1711
Event: time 1516747806.803357, type 3 (EV_ABS), code 1 (ABS_Y), value 12768
Event: time 1516747806.803357, type 3 (EV_ABS), code 24 (ABS_PRESSURE), value 3080
Event: time 1516747806.803357, -------------- SYN_REPORT ------------

// the above is the last event actually reported by hardware when I quickly moved the pen away

Event: time 1516747807.053439, type 4 (EV_MSC), code 4 (MSC_SCAN), value d0042
Event: time 1516747807.053439, type 1 (EV_KEY), code 330 (BTN_TOUCH), value 0
Event: time 1516747807.053439, type 3 (EV_ABS), code 24 (ABS_PRESSURE), value 0
Event: time 1516747807.053439, -------------- SYN_REPORT ------------

// This is the first synthetic event generated by proximity-out emulation code.
// This will not be generated if the tablet behaves well (e.g. reports TOUCH=0 and PRESSURE=0 itself)

Event: time 1516747807.053451, type 1 (EV_KEY), code 320 (BTN_TOOL_PEN), value 0
Event: time 1516747807.053451, -------------- SYN_REPORT ------------

// and this is the second synthetic event generated by proximity-out emulation code
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 obj-m := hid-kye.o hid-uclogic.o hid-polostar.o hid-viewsonic.o
-hid-uclogic-objs := hid-uclogic-core.o hid-uclogic-rdesc.o
+hid-uclogic-objs := hid-uclogic-core.o hid-uclogic-rdesc.o hid-uclogic-proxemu.o
 KVERSION := $(shell uname -r)
 KDIR := /lib/modules/$(KVERSION)/build
 PWD := $(shell pwd)
--- a/hid-uclogic-core.c
+++ b/hid-uclogic-core.c
@@ -19,6 +19,7 @@
 #include <asm/unaligned.h>
 #include "usbhid/usbhid.h"
 #include "hid-uclogic-rdesc.h"
+#include "hid-uclogic-proxemu.h"
 
 #include "hid-ids.h"
 
@@ -59,6 +60,7 @@ struct uclogic_drvdata {
 	bool ignore_pen_usage;
 	bool has_virtual_pad_interface;
 	bool is_hires;
+	struct proxemu_data proxemu;
 };
 
 static __u8 *uclogic_report_fixup(struct hid_device *hdev, __u8 *rdesc,
@@ -401,6 +403,11 @@ static int uclogic_probe_tablet_hires(struct hid_device *hdev,
 	memcpy(drvdata->rdesc, rdesc_template_ptr, drvdata->rsize);
 	uclogic_fill_placeholders(drvdata->rdesc, drvdata->rsize, params, sizeof(params)/sizeof(*params));
 
+	/* Emulate proximity-out events on this type of tablets.
+	 * Hardware proximity-out is not reported due to a firmware bug.
+	 */
+	uclogic_proxemu_init(&drvdata->proxemu, hdev);
+
 	rc = 0;
 
 cleanup:
@@ -569,12 +576,12 @@ static int uclogic_probe(struct hid_device *hdev,
 						hdev,
 						uclogic_rdesc_tablet_template_arr,
 						uclogic_rdesc_tablet_template_size);
+				drvdata->invert_pen_inrange = true;
 			}
 			if (rc) {
 				hid_err(hdev, "tablet enabling failed\n");
 				return rc;
 			}
-			drvdata->invert_pen_inrange = true;
 
 			rc = uclogic_probe_buttons(hdev);
 			drvdata->has_virtual_pad_interface = !rc;
@@ -676,6 +683,17 @@ static int uclogic_probe(struct hid_device *hdev,
 	return 0;
 }
 
+static void uclogic_remove(struct hid_device *hdev)
+{
+	struct uclogic_drvdata *drvdata = hid_get_drvdata(hdev);
+
+	hid_hw_stop(hdev);
+
+	uclogic_proxemu_stop (&drvdata->proxemu);
+
+	devm_kfree(&hdev->dev, drvdata);
+}
+
 #ifdef CONFIG_PM
 static int uclogic_resume(struct hid_device *hdev)
 {
@@ -721,6 +739,8 @@ static int uclogic_raw_event(struct hid_device *hdev, struct hid_report *report,
 		else if (drvdata->invert_pen_inrange)
 			/* Invert the in-range bit */
 			data[1] ^= 0x40;
+
+		uclogic_proxemu_raw_event (&drvdata->proxemu, data, size);
 	}
 
 	return 0;
@@ -759,6 +779,7 @@ static struct hid_driver uclogic_driver = {
 	.name = "uclogic",
 	.id_table = uclogic_devices,
 	.probe = uclogic_probe,
+	.remove = uclogic_remove,
 	.report_fixup = uclogic_report_fixup,
 	.raw_event = uclogic_raw_event,
 	.input_mapping = uclogic_input_mapping,
--- a/hid-uclogic-core.c
+++ b/hid-uclogic-core.c
@@ -217,7 +217,6 @@ static int uclogic_input_configured(struct hid_device *hdev,
 }
 #undef RETURN_SUCCESS
 
-
 /**
  * Enable fully-functional tablet mode and retrieve device parameters.
  *
@@ -494,6 +493,32 @@ cleanup:
 	return rc;
 }
 
+static void uclogic_probe_model (const struct hid_device_id *id, struct hid_device *hdev)
+{
+	int rc;
+	struct usb_device *usb_dev = hid_to_usb_dev(hdev);
+	const char *vendor = NULL;
+	char str_buf [64];
+
+	switch (id->vendor) {
+		case USB_VENDOR_ID_HUION:	vendor = "Huion"; break;
+		case USB_VENDOR_ID_KYE:		vendor = "Genius"; break;
+		case USB_VENDOR_ID_UCLOGIC:	vendor = "UC-Logic"; break;
+		case USB_VENDOR_ID_POLOSTAR:	vendor = "Polostar"; break;
+		case USB_VENDOR_ID_UGTIZER:	vendor = "UGTizer"; break;
+		case USB_VENDOR_ID_UGEE:	vendor = "Ugee"; break;
+	}
+
+	rc = usb_string(usb_dev, 0x79, str_buf, sizeof(str_buf));
+	if (rc >= 0) {
+		if (vendor)
+			snprintf(hdev->name, sizeof(hdev->name), "%s %s Tablet", vendor, str_buf);
+		else
+			snprintf(hdev->name, sizeof(hdev->name), "%s Tablet", str_buf);
+	} else if (vendor)
+		snprintf(hdev->name, sizeof(hdev->name), "%s Tablet", vendor);
+}
+
 static int uclogic_probe(struct hid_device *hdev,
 		const struct hid_device_id *id)
 {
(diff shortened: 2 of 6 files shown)


spbnick:
Thank you for the pull request, Andrey. Unfortunately, I'm traveling for work the past week and failed to find some free time to look over the patches. I'll still be traveling up until Monday and hopefully next week will find the time to do a review. If you're coming to FOSDEM, will be glad to see you there.

anpaza:
ok no problem, the winter in spb is long, I can wait :)

loentar:
@anpaza I tested your fork and found it works with my GAOMON pd1560. Pressure recognized correctly and resolution seems to be improved.
But there is a problem when pointer when it on the very right side of the screen. When pen reaches ~65k by X the cursor jumps to the left side.
it's very similar to the 2byte unsigned integer overflow (x > 65535?)

ScislaC:
@spbnick I tested this on Ubuntu 18.04 with a HUION H640P and it worked as expected. If there's any particular testing you'd like done please let me know. I also couldn't reproduce the issue mentioned by @loentar (granted that issue was reported against a different brand of tablet).

loentar:
To fit higher byte I just divide lower bytes by 2 and place higher bit into freed bit:

```

static int uclogic_raw_event(struct hid_device *hdev, struct hid_report *report,
			u8 *data, int size)
	struct uclogic_drvdata *drvdata = hid_get_drvdata(hdev);

	if ((report->type == HID_INPUT_REPORT) &&
	    (report->id == UCLOGIC_PEN_REPORT_ID || report->id == UCLOGIC_HIRES_PEN_REPORT_ID) &&
	    (size >= 2)) {

        if (drvdata->is_hires && size >= 12) {
            u16* data16 = (u16*)data;
            data16[1] = ((data16[1] >> 1) & 0x7FFF) | (data16[4] << 15);
            data16[2] = ((data16[2] >> 1) & 0x7FFF) | (data16[5] << 15);
            data16[4] = 0;
            data16[5] = 0;
        }

....
```

After that it works perfectly.. but with precision loss of course


Found another issue, now with keyboard.. The first button stopped working, it worked in main repo (but the last didn't worked instead) :) Absolutely no action when pressing under evtest.

loentar:
@anpaza Can you help me with buttons, please? As I understand need to fix some parameters here?

https://github.com/anpaza/digimend-kernel-drivers/blob/244b96926cd48e80875bea7e7844d15070999880/hid-uclogic-rdesc.c#L791

anpaza:
I've looked into the HID standard and found in chapter 5.8 the requirement that multibyte values are stored sequentially in little-endian format. In other words, HID standard does not support the high byte stored separately from the lower bytes. Since the hid-uclogic driver simulates a  HID-compliant device (although chinese tablets are very, very far from being HID compliant), it is impossible to support such reports where the high byte is placed separately.

So, there are two solutions here. One is to scale the multibyte value down so that it fits in 16 bits. The loss of resolution is unsignificant, in my opinion. But the tablet resolution in HID report descriptor also has to be fixed, of course.

Another solution would be to remix the report bytes so that there are two sequential HID-compliant 32-bit values. In this case the tablet resolution in HID report shouldn't be changed, but instead the format report itself should be changed.

I think the second solution is the way to go, but I'm not sure these changes should be done in this pull request. I think it would be wiser first to wait till spbnick mergese this patchset into repository, and then proceed this as a separate patch.

About the 1st button, I don't see what's the problem. In your reports the button bit is clearly seen in its place. Rather than inserting printk's into the driver, you could just do

cat /sys/kernel/debug/hid/<tablet device id>/events

that would also show you what kernel thinks about button states.

loentar:
Thank you very much for the answer!
I didn't know about /sys/.../events, thanks, will try it. `evtest` in my case didn't show anything when I press first button because the event was corrupted.

Regarding non-working first button... False alert.. that was my stupid mistake.. I applied bit shift without filtering by event type, so, button events was changed too.. I spent hours to figure out this stupid problem :laughing: . sad..

More correct but still ugly workaround would be:
```
        if (data[0] == 0x08 && (data[1] & 0xFE) == 0x80 && drvdata->is_hires && size >= 12) {
            // fix tablet
            u16 *data16 = (u16*)data;
            data16[1] = ((data16[1] >> 1) & 0x7FFF) | (data16[4] << 15);
            data16[2] = ((data16[2] >> 1) & 0x7FFF) | (data16[5] << 15);
            data16[4] = 0;
            data16[5] = 0;
        }
```

Do you know any method to extract model info like "GAOMON pd1560" to apply this workaround to specific device only, not to any matched "256c:006e"?

anpaza:
That's why I asked you about the output of uclogic-probe (from [uclogic-tools](https://github.com/DIGImend/uclogic-tools) git repository). It contains some id info, like the string resource with identifier 0x79 (which contains "HA60-F400" on my Huion 1060P), see [this patch](https://github.com/DIGImend/digimend-kernel-drivers/pull/114/commits/ea63deefb56c040aae88e00d7ac9637e76e6b519). It is used to form a more or less unique tablet identifier, e.g. evtest would show the name of your tablets as "Tablet XXX Pad/Pen", where XXX is string resource 0x79.

spbnick:
@anpaza, what would you prefer: to get a review from me, asking to change things, or me to rework this myself into what I think works best? In the first case you might see what kind of code I expect, and get later merges faster, in the second case I get to spend less time and get your code in sooner. If I'm to rework this myself, I'll post a PR and will ask you for a review.

loentar:
Looks like new Huion-based tablets always returns empty string on `0x79` string request. And the `uclogic_probe_model` generates `Huion  Tablet` for those tablets, which does not make any sense. (Also checking for `rc >= 0` isn't correct)

Luckily there are `0xC9` string request which returns OEM device name for those tablets in format: `<VENDOR>_<MODEL>_<VERSION>`. So it's possible to take info there, as fallback.

E.g. for Huion H950P it will be `HUION_T172_170803` -> Vendor=HUION, Model=T172. 
For my Gaomon pd1560 it would be `OEM02_M177_170831` -> Vendor=OEM02, Model=M177. 
Not exactly the model name, but anyway it's much better than some generic `Huion` with empty model name.

Updated version of `uclogic_probe_model` could be something like that:
https://gist.github.com/loentar/939334593a73a11dc2f64bd586b9ad81

I tested it on my pen display and it worked.

anpaza:
@spbnick feel free to rework it yourself, it will be faster than the usual ping-pong.
@loentar I will look into that after Nikolay finishes integrating this patchset.

spbnick:
I spent about 10 hours in total over the last several weeks refactoring the driver and integrating support for the new Huion devices. I'm getting closer to actually testing the code. I found one interesting thing, though: in default mode the tablet *does* send the proximity bit. All four new tablet models I have do that. So, it's a real pity the new high resolution mode doesn't do that.

I'll try to talk to Huion to get them to fix this in newer models at least.

anpaza:
It's a obvious bug, easy to fix. We can keep the artificial proximity-out only for known buggy firmware versions.
Would be cool if it's possible to upgrade firmware via e.g. microSD or USB.

spbnick:
For the start, Huion promised they will fix it in the future models. I will try to talk to them about firmware update and and other issues later, once I get some actual results for them.

spbnick:
@anpaza, is there a particular reason why you used "scheduled work" instead of the timer API (i.e. add_timer/mod_timer)?

I'm getting close to functioning refactoring and am working on the last major piece: proximity emulation. If you're curious here's my WIP: https://github.com/spbnick/digimend-kernel-drivers/tree/huion_hires_refactor

spbnick:
Alright, I implemented the proximity detection emulation with timers. If I did it right, it seems it's much simpler.

spbnick:
Now I need to fixup the frame buttons.

anpaza:
No, there wasn't any special reason to use schedule_delayed_work. If with timers code is simpler, let's use them.

anpaza:
Could you please put something meaningful in hdev->name on tablet detection, just like uclogic_probe_model() does in my patch? Having a device name like "HID 256c:blah" is not very useful, besides, things like libwacom will like to make difference between different tablets with same vendor 0x256c (and libwacom detects tablet model by device->name).

spbnick:
@anpaza, I noticed your `uclogic_probe_model` function, and I appreciate your effort to improve the situation. I'm not sure, however, if this is a fight we want to fight. The manufacturer's behavior regarding VID:PID, manufacturer and product strings were nothing but chaotic.

It would be nice to have good tablet names, but I dread to imagine what such code would turn into over time. I think initialization code would be quite enough of a maintenance burden by itself. I would like to work with wacom driver developers instead to try to make libwacom support generic tablets better. At least that's the plan. If *that* fails, then let's look at trying to make names for our tablets.

anpaza:
Even if libwacom maintainers will agree, still there's the average user case when he runs "xinput list" and sees random garbage names that tells him nothing.
I understand your apprehension, but if you look at almost every driver in linux tree, you will see lots of tables, quirks and workarounds for specific models, so this is not something exclusive here.
Besides, "VID 256c:blah" is not unique, different tablets use the same VID:PID, so the only way to find out the correct tablet model (and libwacom needs the exact model, to choose the right tablet picture, number of buttons etc) is to query the USB device strings.
One more reason, if we change that now, we have a chance to influence the future for that "chaotic" usage of product strings. If we don't, we'll stay with it forever.

spbnick:
Yes, I agree, there's plenty of quirks and workarounds. I'm not sure assigning names to tablets is on par with those, but I'm willing to consider that. The problem is writing and maintaining the heuristics for that needs a lot of work, as some tablets don't have anything much to identify them. No product, no manufacturer strings, no unique VID:PID pairs. All we can do in such cases is use some sort of "fingerprinting" - e.g. matching original report descriptors, parameters retrieved from device, proprietary string descriptors, various other USB descriptors, etc. I would prefer to simply have libwacom handle "unknown" tablets well. If that fails, then we can do the renaming in the kernel.

You can help with making the renaming idea easier to digest and implement, though: contribute tablet information to the [tablet repo](https://github.com/DIGImend/tablets). In particular, `lsusb -v` output and `uclogic-probe` output. Scraping the issues can get us more data, reaching actual tablet owners some more.

anpaza:
Sure, tablet repo was exactly the place where I was looking when I implemented uclogic_probe_model().
My tablet was already there, though.

spbnick:
@anpaza, I think I finally finished all I wanted to do regarding the refactoring and support for the new Huion tablets and would appreciate your review in #134. This still doesn't have the tablet model renaming, but we can discuss that separately again, if you'd like.

anpaza:
This is obsolete now.

PR #125, opened 2018-03-11
Fix building Debian package

danghai:
Hi Nick, 
Pull Request for fixing building Debian package. Because it always creates the `debhelper-build-stamp` file  after the command `dpkg-buildpackage -b -uc`. I put it into gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 .*
 /Module.symvers
 /modules.order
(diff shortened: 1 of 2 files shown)


spbnick:
Thanks, Hai! I pushed this as two separate commits, because the two changes you do here are strictly-speaking unrelated.

Issue #119, opened 2018-02-17
Setup CI


spbnick:
Setup CI with Travis CI. Some ideas of what can be tested:

* Build with `make`
* Install with `make install` and check all modules are installed and can load
* Uninstall with `make install` and check everything is removed
* Build tarball (requires #101)
* Build Debian package
* Install Debian package and check all modules are installed and can load
* Uninstall Debian package and check everything is removed
* Build and install with DKMS and check everything is installed and modules can load (might require a fix to DKMS)
* Remove with DKMS and check everything is removed
* Check versions match across modules, Debian package, DKMS configuration, and everywhere else

PR #126, opened 2018-03-11
Setup CI

danghai:
Hi Nick, 
Pull request for issue: #119

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,27 @@
+language: c
+sudo: required
+compiler:
+  - gcc
+os: linux
+before_install:
+- sudo apt-get -qq update
+- sudo apt-get install -y dpkg-dev debhelper dkms lintian
+- sudo apt-get install linux-headers-`uname -r`
+script:
+- make
+# Getting PACKAGE_VERSION release
+- PACKAGE_VERSION=`. dkms.conf; echo $PACKAGE_VERSION`
+# Testing add, build, install and remove dkms
+- sudo git checkout-index --all --prefix=/usr/src/digimend-$PACKAGE_VERSION/
+- sudo dkms add -m digimend -v $PACKAGE_VERSION
+- sudo dkms build -m digimend -v $PACKAGE_VERSION
+- sudo dkms install -m digimend -v $PACKAGE_VERSION
+- sudo dkms uninstall -m digimend -v $PACKAGE_VERSION
+# Testing Debian package installation, configuration, and uninstall
+- dpkg-buildpackage -b -uc
+# Checks Debian packages for common inconsistencies and errors
+- lintian -i -I --profile debian --show-overrides ../digimend-dkms_*.changes
+- sudo dpkg -i ../digimend-dkms_*_all.deb
+- sudo dpkg -r digimend-dkms
+- sudo make install
+- sudo make uninstall
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 .*
 /Module.symvers
 /modules.order
+!.travis.yml


.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+sudo: required
+dist: trusty
+compiler:
+  - gcc
+os: linux
+before_install:
+- sudo apt-get -qq update
+- sudo apt-get install -y dpkg-dev debhelper dkms lintian
+- sudo apt-get install linux-headers-`uname -r`
+- sudo cp -R . /usr/src/digimend-7
```
spbnick: Could you try replacing this with `git checkout-index --all --prefix /usr/src/digimend-7`? This way we won't copy the whole `.git` directory with the history there. Also, could you perhaps move this to the `script` section? It feels more a part of the tests.

.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+script:
+- make
+- sudo dkms add -m digimend -v 7
+- sudo dkms build -m digimend -v 7
+- sudo dkms install -m digimend -v 7
+- sudo dkms uninstall -m digimend -v 7
```
spbnick: Could you somehow make this work without explicitly mentioning the version number? Either by omitting it altogether or extracting it automatically somehow. This way we won't have to update this file too on release.

.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+- dpkg-buildpackage -b -uc
+- lintian -i -I --profile debian --show-overrides ../digimend-dkms_*_amd64.changes
```
spbnick: Neat that you figured this out, but what does it do? Also, could you please make it not depend on the particular architecture we're building on?

.travis.yml
```diff
@@ -0,0 +1,23 @@
+language: c
+sudo: required
+dist: trusty
```
spbnick: Do we have to mention the distribution here? Can we make it work without that? It would be better not to care about distro updates on Travis CI.

danghai:
Hi Nick, 
Thanks review. I have updated. 
1. I already introduce `$PACKAGE_VERSION ` for `dkms`, and put them under script for testing. 
2. The command `git checkout-index --all --prefix /usr/src/digimend-7` that you recommend to use is missed `/` :(. Fortunately, they mention `The final "/" is important. The exported name is literally just prefixed with the specified string` in man page :)
3. I use the `lintian` because you mention about checking Debian build package.
Thanks

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+  - gcc
+os: linux
+before_install:
+- sudo apt-get -qq update
+- sudo apt-get install -y dpkg-dev debhelper dkms lintian
+- sudo apt-get install linux-headers-`uname -r`
+script:
+- make
+# Getting PACKAGE_VERSION release
+- PACKAGE_VERSION=$(cat dkms.conf | grep  -i  PACKAGE_VERSION= | cut -d '"' -f 2)
```
spbnick: Since dkms is written in shell and simply sources that file, we can use shell
to get the PACKAGE_VERSION, which would be shorter and more reliable:

    PACKAGE_VERSION=`. dkms.conf; echo $PACKAGE_VERSION`

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+# Testing add, build, install and remove dkms
+- sudo git checkout-index --all --prefix=/usr/src/digimend-$PACKAGE_VERSION/
+- sudo dkms add -m digimend -v $PACKAGE_VERSION
+- sudo dkms build -m digimend -v $PACKAGE_VERSION
+- sudo dkms install -m digimend -v $PACKAGE_VERSION
+- sudo dkms uninstall -m digimend -v $PACKAGE_VERSION
+# Testing Debian package installation, configuration, and uninstall
+- dpkg-buildpackage -b -uc
+- lintian -i -I --profile debian --show-overrides ../digimend-dkms_*_`dpkg --print-architecture`.changes
```
spbnick: Could you please add a comment explaining what this line does? Thank you.

spbnick: Also, I think we can simply say `../digimend-dkms_*.changes` here. I don't think we would build more than one architecture, and even if we did, it would be nice to check all of them.

danghai:
I have updated

spbnick:
Thank you very much, Hai, merged!

This is a very good start. Another thing we can do is check that modules can be loaded and unloaded after each installation step, check that modules come from our install location, and not from the stock kernel, and check that versions of our modules match the version we install.

danghai:
Thanks Nick, could you elaborate the test case in new issue?

PR #163, opened 2018-06-25
Update testing dkms in Travis

danghai:
I make PR to update testing `dkms` in Travis. I am not sure the option `--no-depmod`. It seems there is no option for `dkms install`. You can see the error in status Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,14 +9,9 @@ before_install:
 - sudo apt-get install linux-headers-`uname -r`
 script:
 - make
-# Getting PACKAGE_VERSION release
-- PACKAGE_VERSION=`. dkms.conf; echo $PACKAGE_VERSION`
 # Testing add, build, install and remove dkms
-- sudo git checkout-index --all --prefix=/usr/src/digimend-$PACKAGE_VERSION/
-- sudo dkms add -m digimend -v $PACKAGE_VERSION
-- sudo dkms build -m digimend -v $PACKAGE_VERSION
-- sudo dkms install -m digimend -v $PACKAGE_VERSION
-- sudo dkms uninstall -m digimend -v $PACKAGE_VERSION
+- sudo make dkms_install
+- sudo make dkms_uninstall
 # Testing Debian package installation, configuration, and uninstall
 - dpkg-buildpackage -b -uc
 # Checks Debian packages for common inconsistencies and errors
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ dkms_modules_install: dkms_check
 	fi
 	dkms add .
 	dkms build $(DKMS_MODULES)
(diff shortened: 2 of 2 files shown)


spbnick:
This might be because dkms in Ubuntu 14.04.5 LTS is too old. Install it into a vm (e.g. in virt-manager), and check. If that's so, we might need to remove --no-depmod option from the Makefile, or implement detection of dkms versions which don't support it, according to the code complexity vs. performance hit trade-off.

danghai:
@spbnick right, I try to install ubuntu-18.04 , and it works with option `--no-depmod`. My old ubuntu 14 does not work for both `dkms uninstall` and `dkms install`. So what do you think good solution? Could you implement detection of dkms versions? It means write `if-condition` in Makefile to make sure work for old and newest distribution?

spbnick:
First, measure what the time difference for running with and without `--no-depmod` option will be, on 18.04. Use the `time` built-in command, watch the `real` line. I.e. `time sudo make_dkms_install`. If it's just a few seconds, just submit a PR removing the `--no-depmod` option for now.

danghai:
The time: 
```
install hid-rebind /sbin/hid-rebind
install -m 0644 90-hid-rebind.rules /lib/udev/rules.d/90-hid-rebind.rules
udevadm control --reload

real	0m4.218s
```
What does the option `--no-depmod` do?

spbnick:
Read the dkms manual :) Try removing it and compare the installation/uninstallation time.

danghai:
The time without option: `0m4.503s`

spbnick:
And with the option? Is the difference worth having the detection code added?

danghai:
In Ubuntu 18:
with `--no-depmod`: 0m4.218s
without  `--no-depmod`: 0m4.503s
In ubuuntu 14:
without  `--no-depmod`: 0m16.102s

spbnick:
Really? No difference with `--no-depmod`? Wow. OK, could you please post a PR removing the `--no-depmod` use both from `dkms_install` and `dkms_uninstall`. Thank you.

danghai:
@spbnick yeah, I have checked 2 times. There is no much difference time between with/without `--no-depmod`. I add 1 commit to remove the option `--no-depmod`. Travis is frozen now :)

spbnick:
Thank you, Hai! I switched the order of commits to keep the CI working, added commit descriptions and fixed one comment, in PR #166. Closing this one.

danghai:
@spbnick Could you help me how to reorder the commit in git quickly? For example, if I have 3 commits `A --> B --> C`, How can I reorder commit `C-->B-->A`?

spbnick:
Use interactive rebase (`git rebase -i`), and then just switch the lines around :)

danghai:
Oh Thanks. It is really simple :)

PR #237, opened 2018-12-09
Simplified Fedora install

BrunoVernay:
It works the same way as Debian.  I tested it on my Distro (Fedora 29).

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ your kernel installed (on Debian-based systems):
 
 or (on Fedora-based systems):
 
-    sudo dnf install -y "kernel-devel-uname-r == $(uname -r)"
+    sudo dnf install -y "kernel-devel-$(uname -r)"
 
 If you get "Error: Unable to find a match" from the above command, make sure
 your kernel is up-to-date, and if not, update it and try again.


spbnick:
Thank you, Bruno! I thought the point of using `kernel-devel-uname-r` is that it will get upgraded, thus pulling the fresh `kernel-devel-$(uname -r)` package with it. Whereas if you directly install the package with the version in the name, that will stay forever. Am I missing something?

BrunoVernay:
I made a little research, but did not find a clear explanation.
The only reason I would see is to allow to use:
```
> sudo dnf install kernel-devel-uname-r                
Last metadata expiration check: 1:38:47 ago on dim. 09 déc. 2018 18:09:33 CET.
Package kernel-devel-4.19.4-300.fc29.x86_64 is already installed.
Package kernel-devel-4.19.5-300.fc29.x86_64 is already installed.
Package kernel-devel-4.19.6-300.fc29.x86_64 is already installed.
Dependencies resolved.
Nothing to do.
Complete!
```
This is simple and gets the `devel` for all the installed versions at once.  
But I prefer the simplicity and absence of magic ...

vasilvestre:
Could get merged no ?

MrGibus:
One thing to add for fedora issues is that devices [aren't always recognised](https://github.com/xournalpp/xournalpp/issues/2273) and you need to `unset GDK_CORE_DEVICE_EVENTS` to fix it.

PR #290, opened 2019-06-18
Remove redudant check of ver_ptr

danghai:
Hi Nick,

I make PR to remove redundant check of ver_ptr


spbnick:
Ah, likely a copy-paste mistake. Thank you, Hai :)

Issue #388, opened 2020-05-04
XP-PEN Star G430S_B doesn't work with DIGImend


Sturwandan:
This tablet is similar in appearance to the white version has same size and buttons, however, USB ID is different.
Black tablet ID is 28bd:0913
White tablet ID is 28bd:0075

When I insert the tablet, following messages appear:
```
usb 6-3: new full-speed USB device number 37 using ohci-pci
usb 6-3: New USB device found, idVendor=28bd, idProduct=0913, bcdDevice= 0.00
usb 6-3: New USB device strings: Mfr=1, Product=2, SerialNumber=3
usb 6-3: Product: 4 inch PenTablet
usb 6-3: Manufacturer: UGTABLET
usb 6-3: SerialNumber: 000000
input: UGTABLET 4 inch PenTablet Mouse as /devices/pci0000:00/0000:00:16.0/usb6/6-3/6-3:1.0/0003:28BD:0913.0085/input/input151
input: UGTABLET 4 inch PenTablet Keyboard as /devices/pci0000:00/0000:00:16.0/usb6/6-3/6-3:1.0/0003:28BD:0913.0085/input/input152
hid-generic 0003:28BD:0913.0085: input,hidraw0: USB HID v1.00 Mouse [UGTABLET 4 inch PenTablet] on usb-0000:00:16.0-3/input0
input: UGTABLET 4 inch PenTablet as /devices/pci0000:00/0000:00:16.0/usb6/6-3/6-3:1.1/0003:28BD:0913.0086/input/input153
hid-generic 0003:28BD:0913.0086: input,hidraw1: USB HID v1.00 Device [UGTABLET 4 inch PenTablet] on usb-0000:00:16.0-3/input1
hid-generic 0003:28BD:0913.0087: hiddev96,hidraw2: USB HID v1.00 Device [UGTABLET 4 inch PenTablet] on usb-0000:00:16.0-3/input2
```

The tablet works with generic drivers, pressure sensitivity is available in GIMP, however, buttons on the pen do not work as expected: the button next to the tip works as MMB (middle mouse button), while the button further from the tip doesn't work at all. Expected behavior is: close button: RMB, far button: MMB

The far button works in principle, since this command ``usbhid-dump -m 28bd:0913 -e stream`` shows lines like ``006:036:001:STREAM             1588586846.485230
 07 A0 E0 3B 0C 22 00 00 00 00`` when no buttons are pressed, but 
[...]

Issue #351, opened 2020-02-12
Support for XP-Pen Artist Pro 12


Alex-PK:
First of all: thank you all for your work!

I just got an XP-Pen Artist Pro 12, and it's working with the standard drivers, but the buttons are not configurable.

```
$ lsusb 
Bus 001 Device 012: ID 28bd:1227 UGTABLET TABLET

$ xinput
⎡ Virtual core pointer                    	id=2	[master pointer  (3)]
⎜   ↳ Virtual core XTEST pointer              	id=4	[slave  pointer  (2)]
⎜   ↳ Logitech K270                           	id=9	[slave  pointer  (2)]
⎜   ↳ Logitech M185                           	id=10	[slave  pointer  (2)]
⎜   ↳ Logitech M525                           	id=11	[slave  pointer  (2)]
⎜   ↳ HOLTEK USB-HID Keyboard Mouse           	id=17	[slave  pointer  (2)]
⎜   ↳ HOLTEK USB-HID Keyboard Consumer Control	id=19	[slave  pointer  (2)]
⎜   ↳ UGTABLET TABLET Mouse                   	id=14	[slave  pointer  (2)]
⎣ Virtual core keyboard                   	id=3	[master keyboard (2)]
    ↳ Virtual core XTEST keyboard             	id=5	[slave  keyboard (3)]
    ↳ Power Button                            	id=6	[slave  keyboard (3)]
    ↳ Power Button                            	id=7	[slave  keyboard (3)]
    ↳ Sleep Button                            	id=8	[slave  keyboard (3)]
    ↳ Dell Dell AC511 USB SoundBar            	id=12	[slave  keyboard (3)]
    ↳ HOLTEK USB-HID Keyboard                 	id=16	[slave  keyboard (3)]
    ↳ HOLTEK USB-HID Keyboard System Control  	id=18	[slave  keyboard (3)]
    ↳ HOLTEK USB-HID Keyboard                 	id=20	[slave  keyboard (3)]
    ↳ Logitech K270                           	id=21	[slave  keyboard (3)]
    ↳ HOLTEK USB-HID Keyboard Consumer Control	id=22	[slave  keyboard (3)]
    ↳ UGTABLET TABLET Keyboard                	id=13	[slave  keyboard (3)]
    ↳ UGTABLET TABLET                         	id=15	[slave  keyboard (3)]

$ ./uclogic-probe 1 2
[...]

PR #397, opened 2020-05-21
A driver for XP-PEN G430S_B (black) tablet

Sturwandan:
This driver works with my tablet with one problem:
The tablet only works until it's reconnected, then driver needs to be reinserted for the tablet to work again.

I wasn't able to pinpoint this problem yet. However simpler driver from #388 does not have such a problem.

### DIFF
--- a/hid-g430sb.c
+++ b/hid-g430sb.c
@@ -1,0 +1,230 @@
+// SPDX-License-Identifier: GPL-2.0+
+/*
+ *  HID driver for XP-PEN G430S_B (black) tablet, which is not fully compliant with HID standard
+ *  Based on hid-viewsonic.c by Copyright (c) 2017 Nikolai Kondrashov
+ *  Version: 1.1
+ *
+ *  Copyright (c) 2020 Sturwandan
+ */
+
+/*
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; either version 2 of the License, or (at your option)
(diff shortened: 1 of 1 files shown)


Sturwandan:
I have forgotten to add patch to all other files. I don't know how to make github accept a patch file. 
https://github.com/DIGImend/digimend-kernel-drivers/issues/388#issuecomment-627261183
However this patch is trivial and same for simple and advanced version of the driver.

alfazaz:
I found your work and it's interesting... I'm trying also to make another XP PEN tablet to work here (without success... init is ok, usbhid-dump is ok but no events are detected by evtest :-( ) :
https://github.com/DIGImend/digimend-kernel-drivers/issues/351

I based my (not working) work on the following link. It's another XP PEN Artist Pull Request which is not structured like your Pull Request : 
https://github.com/DIGImend/digimend-kernel-drivers/pull/291

May be it can inspire you (on reconnect problem or other things ?) even if I can't get it work for my XP PEN tablet model.

Sturwandan:
Check the last comment in my bug report ( #388 ), I have had the same problem. The reason was that when tablet is in native mode (after I sent init), the events are emitted from interface 2 (starting from 0), rather than from interface 1 (which they are sent from in default mode). I have spent quite some time trying to debug it, until I got a suggestion to make sure I am using the correct interface.

alfazaz:
Thanks @Sturwandan . I answered in https://github.com/DIGImend/digimend-kernel-drivers/issues/388 .

PR #483, opened 2020-11-19
Fix for the Artisul D16

beyondoblivion:


kurikaesu:
Does this still work correctly? If so, can you squash all the commits into a single commit?

beyondoblivion:
It still works correctly for me. I'm still not sure how to merge all the commits into one. I'm trying to figuring it out

PR #553, opened 2021-08-04
Use HID_QUIRK_HIDINPUT_FORCE to fix #550

### LINKED ISSUE #550: XP-PEN G430S works in v9 -- broken in v10 by #464
I have a XP-PEN G430S which works perfectly with v9 but doesn't work at all with v10.

I tracked the issue to #464 (commit ec5c16d631940546ddaf03eecec082865bd9c676).

Indeed, with the following patch applied the pen works fine.
```diff
--- a/hid-uclogic-rdesc.c
+++ b/hid-uclogic-rdesc.c
@@ -534,7 +534,7 @@ const size_t uclogic_rdesc_twha60_fixed1_size =
 /* Fixed report descriptor template for (tweaked) v1 pen reports */
 const __u8 uclogic_rdesc_v1_pen_template_arr[] = {
        0x05, 0x0D,             /*  Usage Page (Digitizer),                 */
-       0x09, 0x01,             /*  Usage (Digitizer),                      */
+       0x09, 0x02,             /*  Usage (Digitizer),                      */
        0xA1, 0x01,             /*  Collection (Application),               */
        0x85, 0x07,             /*      Report ID (7),                      */
        0x09, 0x20,             /*      Usage (Stylus),                     */
```

I'm not sure what would be the proper fix for this issue.

Note that G430S has the same usbid as G540 and it's handled by `USB_DEVICE_ID_UGEE_XPPEN_TABLET_G540`.

tornaria:
We must force HIDINPUT, since after ec5c16d some devices are of class `HID_DG_DIGITIZER`, which is (incorrectly?) not considered as input by macro `IS_INPUT_APPLICATION()`.

With this quirk all devices are forced to go through `hidinput_connect()` to be configured as input devices.

For instance, XP-PEN G430S was broken by ec5c16d and it works now.

I also included a second (cosmetic) patch which makes devices of class `HID_DG_DIGITIZER` to use suffix "Digitizer" in the name of the device (used for dmesg, libinput, xinput,...)

### DIFF
--- a/hid-uclogic-core.c
+++ b/hid-uclogic-core.c
@@ -163,6 +163,9 @@ static int uclogic_input_configured(struct hid_device *hdev,
 		case HID_GD_KEYPAD:
 			suffix = "Pad";
 			break;
+		case HID_DG_DIGITIZER:
+			suffix = "Digitizer";
+			break;
 		case HID_DG_PEN:
 			suffix = "Pen";
 			break;
--- a/hid-uclogic-core.c
+++ b/hid-uclogic-core.c
@@ -203,6 +203,14 @@ static int uclogic_probe(struct hid_device *hdev,
 #ifdef HID_QUIRK_NO_EMPTY_INPUT
 	hdev->quirks |= HID_QUIRK_NO_EMPTY_INPUT;
 #endif
+	/*
+	 * We must force HIDINPUT, since some devices are of class
+	 * HID_DG_DIGITIZER, which is (incorrectly?) not considered as
+	 * input by macro IS_INPUT_APPLICATION().
+	 * With this quirk all devices are forced to go through
+	 * hidinput_connect() to be configured as input devices.
+	 */
+	hdev->quirks |= HID_QUIRK_HIDINPUT_FORCE;
 
 	/* Allocate and assign driver data */
 	drvdata = devm_kzalloc(&hdev->dev, sizeof(*drvdata), GFP_KERNEL);


nic3-14159:
Here's some background to what this patch fixes. There seems to be a few XP-PEN tablets (confirmed on G430s and STAR 06)  that no longer work with the DIGImend drivers after commit ec5c16d. After some investigation, we believe the source of the issue is the IS_INPUT_APPLICATION macro in include/linux/hid.h in the kernel tree. 
```
/* Applications from HID Usage Tables 4/8/99 Version 1.1 */
/* We ignore a few input applications that are not widely used */
#define IS_INPUT_APPLICATION(a) \
		(((a >= HID_UP_GENDESK) && (a <= HID_GD_MULTIAXIS)) \
		|| ((a >= HID_DG_PEN) && (a <= HID_DG_WHITEBOARD)) \
		|| (a == HID_GD_SYSTEM_CONTROL) || (a == HID_CP_CONSUMER_CONTROL) \
		|| (a == HID_GD_WIRELESS_RADIO_CTLS))
```
By changing the HID descriptor usage from 0x02 (Pen) to 0x01 (Digitizer), the macro evaluates to false, as the digitizer usage leads to HID_DG_DIGITIZER being passed to the macro, which does not pass `a >= HID_DG_PEN`. 

This macro is used in drivers/hid/hid-input.c, in the function hidinput_connect():
```
if (!force) {
	for (i = 0; i < hid->maxcollection; i++) {
		struct hid_collection *col = &hid->collection[i];
		if (col->type == HID_COLLECTION_APPLICATION ||
				col->type == HID_COLLECTION_PHYSICAL)
			if (IS_INPUT_APPLICATION(col->usage))
				break;
	}

	if (i == hid->maxcollection)
		return -1;
}
```
We believe the reason it only seems to be affecting XP-PEN tablets is that the hid-uclogic-params.c code does not add a HID descriptor for the frame on these models, either because the tablet does not have frame buttons or the frame is not currently supported. The frame descriptors use the Keypad usage from the Generic Desktop page, which passes the IS_INPUT_APPLICATION macro. It seems like the above code snippet passes if at least one usage passes the macro, so without the frame descriptor, the function returns early and does not configure the input layer. Based on this, it seems likely that any tablet with a fixed descriptor that does not contain a usage which passes the macro will also experience similar issues.

Using HID_QUIRK_HIDINPUT_FORCE as suggested by @tornaria leads to the force flag being set to true, bypassing the check. A more permanant fix would be to change HID_DG_PEN to HID_DG_DIGITIZER in the IS_INPUT_APPLICATION macro, as the assumption that the digitizer input application is not widely used no longer seems to be correct after the usages were changed in the DIGImend code.

Issue #446, opened 2020-08-12
Add support for XP-Pen Star G960


jmoy:
I would like to add support for the [XP-Pen Start G960](https://www.xp-pen.com/product/508.html), VID 0x28bd, PID: 0x0920.
This is a tablet with 4 keys on the frame.

I'm attaching the files suggested by the project side.

I've been trying to decipher the reports from the device. The format seems to be:

1. Flags byte
3. 16 bits little-endian Y coordinate (max 32767)
4. 16 bits little-endian X coordinate (max 32767)
5. 16 bits little endian pressure (max 8191)

I've not been able to fully decipher the flag byte. The first 4 bits are either hex 'c' if the stylus is out of range or hex 'a' if the stylus is in range. The next four bits consists of an OR of 1 for stylus touching the tablet and 2 for stylus button pressed.

I know C and fighting with this tablet has given me some idea of how USB works and I'd be ready to work on adding support for this but would appreciate any pointers on how to start.

[descriptors.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/5062784/descriptors.txt)
[hid_report_descriptors.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/5062785/hid_report_descriptors.txt)
[key1.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/5062786/key1.txt)
[key2.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/5062787/key2.txt)
[key3.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/5062788/key3.txt)
[key4.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/5062789/key4.txt)
[pen_buttons.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/5062791/pen_buttons.txt)
[pen_coords.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/5062792/pen_coords.txt)
[pen_pressure.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/5062793/pen_pressure.txt)
[probe.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/5062794/probe.txt)

kurikaesu:
Hi jmoy, I'm following up adding support for your device in my PR here: https://github.com/DIGImend/digimend-kernel-drivers/pull/557

I need you to push the special handshake key to interface 1 then resubmit the data for:
- Pen coordinates
[...]

PR #557, opened 2021-08-19
XP-Pen Deco Pro SM, Artist 12, 13.3, 15.6 and 22R Pro Support

kurikaesu:
Adds some support for the XP-Pen Artist 22R Pro.

I used a bunch of code from PR #291 to get the digitizer/stylus working however the frame buttons are not configured correctly.
Some of the buttons will respond to touches but they do not have unique codes being emitted.

Left button 8 and right button 6 for example emit the key 123 which on my system increases sound volume.

There is still a bunch of test rdescs that I wrote for interfaces 0 and 1 but those interfaces will result in the buttons emitting their default hard coded keyboard keycodes. They will be removed in future.


kurikaesu:
I tested these changes on my XP-Pen 13.3 Pro to see if at least the pen would work and it doesn't.

Moving the pen around results in input irq status -75 received so for the 13.3 Pro the report descriptor for the stylus is wrong.
The frame buttons also do not report any keypresses when using "xinput test"

kurikaesu:
Putting information that I dig up about this tablet here: https://github.com/kurikaesu/DIGImend-research/blob/main/xp-pen-22r-pro.md

Also digging up info on my Artist 13.3 Pro here: https://github.com/kurikaesu/DIGImend-research/blob/main/xp-pen-13.3-pro.md

kurikaesu:
Trying to create a manual report descriptor for the frame keys referring to the usb packet dumps I made but not having much luck. So far there are no IRQ errors in dmesg and the 6th button on the right reports `motion a[3]=-60` when pressed but every other key or dial turn produces nothing.

kurikaesu:
Using the bit of research I posted above I found what was different on the 13.3 pro. The stylus now works on the 13.3 pro with pressure sensitivity and I would guess tilt support.

kurikaesu:
With that last commit it appears that the 13.3 pro's frame / express keys are now working correctly.

```
> xinput list
⎡ Virtual core pointer                    	id=2	[master pointer  (3)]
⎜   ↳ Virtual core XTEST pointer              	id=4	[slave  pointer  (2)]
⎜   ↳ LingYao ShangHai Thumb Keyboard         	id=10	[slave  pointer  (2)]
⎜   ↳ UGTABLET 13.3 inch PenDisplay Pad pad   	id=11	[slave  pointer  (2)]
⎜   ↳ UGTABLET 13.3 inch PenDisplay Pen stylus	id=12	[slave  pointer  (2)]
⎜   ↳ Convertible 2 TKL Consumer Control      	id=14	[slave  pointer  (2)]
⎜   ↳ SteelSeries SteelSeries Rival 3 Wireless Consumer Control	id=20	[slave  pointer  (2)]
⎜   ↳ SteelSeries SteelSeries Rival 3 Wireless	id=22	[slave  pointer  (2)]
⎣ Virtual core keyboard                   	id=3	[master keyboard (2)]
    ↳ Virtual core XTEST keyboard             	id=5	[slave  keyboard (3)]
    ↳ Power Button                            	id=6	[slave  keyboard (3)]
    ↳ Power Button                            	id=7	[slave  keyboard (3)]
    ↳ Microsoft® LifeCam HD-3000: Mi         	id=17	[slave  keyboard (3)]
    ↳ UVC Camera (046d:0825)                  	id=18	[slave  keyboard (3)]
    ↳ Samson Technologies Samson Go Mic       	id=8	[slave  keyboard (3)]
    ↳ LingYao ShangHai Thumb Keyboard         	id=9	[slave  keyboard (3)]
    ↳ Convertible 2 TKL System Control        	id=13	[slave  keyboard (3)]
    ↳ Convertible 2 TKL Consumer Control      	id=15	[slave  keyboard (3)]
    ↳ Convertible 2 TKL                       	id=16	[slave  keyboard (3)]
    ↳ SteelSeries SteelSeries Rival 3 Wireless Keyboard	id=19	[slave  keyboard (3)]
    ↳ SteelSeries SteelSeries Rival 3 Wireless Consumer Control	id=21	[slave  keyboard (3)]
```

```
> xsetwacom list
UGTABLET 13.3 inch PenDisplay Pad pad	id: 11	type: PAD       
UGTABLET 13.3 inch PenDisplay Pen stylus	id: 12	type: STYLUS   
```

```
> xsetwacom set "UGTABLET 13.3 inch PenDisplay Pad pad" Button 1 "key 5"

* press top-most express key button *

5
```

kurikaesu:
Just checked out my Art Deco Pro and it looks like it may be almost identical in implementation as the 13.3 pro. It might be more close to the implementation in #291 actually.

kurikaesu:
For the Artist 22R Pro using the code changes in the PR:

Creating hwdb rule in `/etc/udev/hwdb.d/70-xppen.hwdb` with the contents:
```
id-input:*:input:b0003v28BDp091Be0100*
 ID_INPUT_MOUSE=0
 ID_INPUT_TABLET=1
 ID_INPUT_TABLET_PAD=1
```
Then executing a `sudo udevadm hwdb --update` followed by a  `sudo udevadm trigger /sys/class/input/eventX` where `eventX` should be replaced with whatever `sudo libinput record` says the Pad event id is appears to make the left side buttons (but not the dial) work:

```
> xsetwacom list
UGTABLET 21.5 inch PenDisplay Pen stylus	id: 19	type: STYLUS    
UGTABLET 21.5 inch PenDisplay Pad pad	id: 20	type: PAD 
```

```
> xsetwacom -s get "UGTABLET 21.5 inch PenDisplay Pad pad" all
Property 'Wacom Tablet Area' does not exist on device.
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "Button" "1" "button +1 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "Button" "2" "button +2 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "Button" "3" "button +3 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "Button" "8" "button +8 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "Button" "9" "button +9 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "Button" "10" "button +10 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "Button" "11" "button +11 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "Button" "12" "button +12 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "Button" "13" "button +13 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "Button" "14" "button +14 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "ToolDebugLevel" "0"
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "TabletDebugLevel" "0"
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "Suppress" "2"
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "RawSample" "4"
Property 'Wacom Pressurecurve' does not exist on device.
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "Mode" "Absolute"
Property 'Wacom Hover Click' does not exist on device.
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "Touch" "off"
Property 'Wacom Hardware Touch Switch' does not exist on device.
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "Gesture" "off"
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "ZoomDistance" "0"
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "ScrollDistance" "0"
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "TapTime" "250"
Property 'Wacom Proximity Threshold' does not exist on device.
Property 'Wacom Rotation' does not exist on device.
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "RelWheelUp" "1" "button +5 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "RelWheelDown" "2" "button +4 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "AbsWheelUp" "3" "button +4 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "AbsWheelDown" "4" "button +5 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "AbsWheel2Up" "5" "button +4 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "AbsWheel2Down" "6" "button +5 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "StripLeftUp" "1" "button +4 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "StripLeftDown" "2" "button +5 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "StripRightUp" "3" "button +4 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "StripRightDown" "4" "button +5 "
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "Threshold" "0"
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "BindToSerial" "0"
Property 'Wacom Pressure Recalibration' does not exist on device.
xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" "PanScrollThreshold" "13"
```
```
> xsetwacom set "UGTABLET 21.5 inch PenDisplay Pad pad" Button 1 "key m"

* press top left-most button / express key *
m
```

I suspect the the Deco Pro SM will also start working if the appropriate hwdb entry is created.

Another problem that I don't know how to deal with right now is that there may be only 10 buttons actually defined:
https://github.com/torvalds/linux/blob/master/include/uapi/linux/input-event-codes.h#L342

For the Artist 22R Pro which has 20 buttons and 2 dials with 4 "buttons" (left & right turns emit buttons) I don't know if these can actually be properly mapped.

The Deco Pro SM also has the touch-pad in the dial that also functions as a second dial which brings the total buttons on it to 12. Maybe these dials can be somehow mapped to the StripLeftUp/Down and StripRightUp/Down?

kurikaesu:
After messing around with things more on my Artist 22R Pro, I've been able to do the following:

Assigned named buttons to the right side express keys and both dials:
`/etc/udev/hwdb.d/70-xppen.hwdb`
```
evdev:input:b0003v28BDp091Be0100*
 KEYBOARD_KEY_9000b=0x120
 KEYBOARD_KEY_9000c=0x121
 KEYBOARD_KEY_9000d=0x122
 KEYBOARD_KEY_9000e=0x123
 KEYBOARD_KEY_9000f=0x124
 KEYBOARD_KEY_90010=0x125
 KEYBOARD_KEY_90011=0x126
 KEYBOARD_KEY_90012=0x127
 KEYBOARD_KEY_90013=0x128
 KEYBOARD_KEY_90014=0x129
 KEYBOARD_KEY_90015=0x111
 KEYBOARD_KEY_90016=0x110
 KEYBOARD_KEY_90017=0x115
 KEYBOARD_KEY_90018=0x116
```
Then using Key Mapper: https://github.com/sezanzeb/key-mapper
Map these keys to shortcuts or even macros.

Still a work-around but its better than nothing. If there was a way to map the express keys to the wacom driver button 4, 5, 6, etc. which don't seem to correspond to anything then I would definitely like to know.

nic3-14159:
@kurikaesu Wacom buttons 4-7 are reserved and cannot be mapped to anything. From the digimend readme:
"Note that buttons are numbered 1, 2, 3, 8, 9, 10, and so on, i.e. buttons 4, 5, 6, and 7 are not used. They're reserved for vertical and horizontal scrolling events by the X server."

kurikaesu:
Oh I see! Thanks for the clarification :smiley:
Guess there is naught to do but use tools to remap the keys. At least now they aren't hard-coded keyboard scan-codes.

The 22R Pro's dials don't send a "key-up/stop rotate" state unlike the 13.3 pro so `evtest` or `libinput record` discard the pressed state past the first message. Need to press something else or turn the dial the other way in order to be able to reset the state. Maybe the dials should not be set as buttons on the 22R Pro?

kurikaesu:
Added this to my `/etc/udev/hwdb.d/70-xppen.hwdb`:

```
id-input:*:input:b0003v28BDp0909e0100*
 ID_INPUT_TABLET=1
 ID_INPUT_TABLET_PAD=1
```

This enables the Deco Pro SM:

```
> xsetwacom list
9 inch PenTablet Pen stylus     	id: 19	type: STYLUS    
9 inch PenTablet Pad pad        	id: 22	type: PAD  
``

The Deco Pro SM dials have the same behavior as the Artist 22R Pro so its probably not a good idea to bind a shortcut to them right now.

**Comment by Korvox** [comment]
If you ever want to get these patches properly merged into mainline along with mine, hit me up some time, I think we have to bypass Digimend. 

And tilt support works? I assume the chip is accounting for the tilt in the reported coordinates? Lucky, I guess they learned from the 15.6 that tilt reporting angle but not just hardware-side including that in the position reporting just sucked.

We also might be able to get the Artist 12 Pro included if @alfazaz's work in #351 is applicable.

**Comment by nic3-14159** [comment]
@Korvox I also have some code on my machine for the Star 06 (There is currently pen support but no frame button support, that tablet also needs a "key packet" sent to it) based on your patches, but I was sort of waiting until your patch got merged into DIGImend before posting.

**Comment by kurikaesu** [comment]
@Korvox I haven't tested tilt yet since I don't have a tool test it with nor do I know how to set up tilt in Krita.
I do think that I'll need to merge your tilt fixes in your PR for most if not all of the cases.

Definitely keen on getting these changes all through to mainline once I can deal with the frame dials.
I'll take a look through https://github.com/DIGImend/digimend-kernel-drivers/issues/351 to see what I can use or adapt. I've also reached out to @Deevad in https://github.com/DIGImend/digimend-kernel-drivers/issues/535 to see if I'm able to get the 24 Pro done too.

@nic3-14159 if you fork this repository and push your changes to a branch, link it here and I can also take a look at it and see how we can merge it all together.

**Comment by nic3-14159** [comment]
@Korvox @kurikaesu I'm assuming mainline is referring to the kernel tree? If so, I feel like we should communicate our intentions to @spbnick before trying to bypass DIGImend

**Comment by kurikaesu** [comment]
Last commit changes the Deco Pro's dial to emulate a Wheel (0x38) which overcomes the not resetting state back to 0 issue. Unfortunately we can only have 1 wheel in the report ID so the inner touch dial is now a Dial (0x37).

Both of the dials emit -1 (or -2 for some reason) and 1 indicating positive and negative however I don't think Dial can be bound by Key Mapper for example since it has no idea what a Dial is. Not too sure if dials can also be udev/hwdb rebound to something else either.

**Comment by kurikaesu** [comment]
@nic3-14159 Agreed, for now I'm just happy enough to have my hardware work in Linux.

File: hid-uclogic-core.c [hunk 1]
```diff
@@ -414,8 +414,27 @@ static int uclogic_raw_event_frame(
...
 	if (frame->bitmap_dial_byte > 0 && frame->bitmap_dial_byte < size) {
-		if (data[frame->bitmap_dial_byte] == 2) {
+		switch (data[frame->bitmap_dial_byte]) {
+		case 2:
 			data[frame->bitmap_dial_byte] = -1;
+			break;
+
+		/* Everything below here is for tablets that shove multiple dials into 1 byte */
+		case 4:
+			data[frame->bitmap_dial_byte] = 1;
```
kurikaesu, line 424: Mad brain fart moment. I should just squash these down and allow fall-through in the cases.

kurikaesu:
Looked through https://github.com/DIGImend/digimend-kernel-drivers/issues/351. The 12 Pro and the 13.3 Pro are identical.

Committed the hardware ID to my fork & branch to enable it.

alfazaz:
@kurikaesu Your commit for the 12 Pro works for me... I think it solves #351.

Thanks !

Korvox:
> I don't have a tool test it with nor do I know how to set up tilt in Krita.

Krita natively uses tilt, you can verify its present at least with libinput via `libinput record` (assuming that is the userspace input driver you are using with the device, of course).

Besides the packet function taking different magic values, is there anything else to be made generic? I rebased and amended my branch with what should be something to just base all other devices off of.

alfazaz:
@kurikaesu @Korvox I saw that tilt also works on my 12 Pro with your work. :-)

nic3-14159:
@Korvox I think the interface that should be used to send the init packet and receive reports needs to be general. As I've said before, I believe there are other tablets that I suspect need an init packet which seem to have a different number of interfaces. I'll see if I can dig up relevant issues. I don't believe they were XP-Pen though, so it is possible that they require many other changes. But so far, all XP-Pen tablets I've seen use interface 2. I'll also have to check my Star 06 code does anything majorly different in the params inkt code. (I'm currently away from my main system, I'll have to check later)

nic3-14159:
@Korvox Found it, the Star G960 (#446) reports keyboard events for the frame and has an OUT endpoint on interface 1 (only 2 interfaces). The HID descriptor on interface 1 also appears to be very similar to the interface 2 descriptors on tablets known to need the init packet. All of these are characteristic of tablets needing the init packet, so it does appear that the code needs to support different interfaces in order to support that tablet.

kurikaesu:
Last commit makes at least the left dial on the 22R Pro usable.

@nic3-14159 I should be able to handle the G960, just need to update the python script to inject the key on interface 1 and dump the packets that way. I'll make a comment in https://github.com/DIGImend/digimend-kernel-drivers/issues/446 with the updated script.

As far as different activation keys are concerned, I originally thought my 22R would have a different key so I wrote the function calls with passing of a custom key in mind. I can certainly re-do this, just need to know the second activation key to handle.

kurikaesu:
@Korvox I pulled in your changes into this PR. If the other tablets need tilt compensation then the calculations or values need to be updated for each of them. I tried on a whim applying the compensation as-is to my 22R and it just messed up the pen coordinates in sections of the screen.

Korvox:
> @Korvox I pulled in your changes into this PR. If the other tablets need tilt compensation then the calculations or values need to be updated for each of them. I tried on a whim applying the compensation as-is to my 22R and it just messed up the pen coordinates in sections of the screen.

See this python script I wrote [here](https://github.com/DIGImend/digimend-kernel-drivers/pull/291#issuecomment-526455118). To generate a new tangent table. You basically figure out the height of the sensor (should be the same on all the pro devices since they use the same pen, but differences in material in the tablet screen can vary this slightly relative to the digitizer) and the pixels per inch resolution of the touch points (ie, not the devices physical resolution, the resolution of the touch area).

And do make sure it needs it for tilt to work. If you point the pen in different directions and the point follows your pointing angle than tilt is already being included in the coordinates.

Korvox:
> @Korvox I think the interface that should be used to send the init packet and receive reports might need to be general. As I've said before, I think there are other tablets that I suspect need an init packet which seem to have a different number of interfaces (not 100% sure though). I'll see if I can dig up relevant issues. I don't believe they were XP-Pen though, so it is possible that they require many other changes. But so far, all XP-Pen tablets I've seen use interface 2. I'll also have to check my Star 06 code does anything majorly different in the params init function. (I currently don't have access to my main system, so I'll post my code and check some other things later)

Latest commit adds the interface id as an arg to probe.

nic3-14159:
@kurikaesu @Korvox Do any of you think it may be more efficient if we all just open pulls into one person's fork and then have a single pull to "upstream" DIGImend or something like that? I think that would make it easier for us to collaborate if we want to split up the tasks of supporting more of the tablets that need the init packet. It might be hard to do that now that there are already pull requests with a bunch of discussions, but I feel like things are starting to get a little messy with different pulls/forks somewhat depending on each other and multiple tablets being put into one pull request.

kurikaesu:
There is nothing stopping us from merging another's branch/mainline into our own through git without requiring creating a pull request between us. What I think is more important right now though is to be able to freely experiment with ideas on how to implement the tablets without clobbering each other's work or having to frequently rebase each other's branches or forks which is where having multiple PRs and discussions is fine in my opinion.

When we are done trying things out we can come together and do a proper merge and squash all the commits and close whichever PR we no longer need/want.

nic3-14159:
Alright, I can work with that. I'm also working on a more general handshake script which should automatically detect the correct interface and accept the USB VID:PID as an argument, which should make asking people for dumps a little less work. [(Project here)](https://github.com/nic3-14159/uclogic-init-v3) I'm planning to make it try both the known init keys, as well as properly respect the max USB packet size that the device reports.

I started it in C, but I'm currently rewriting it in python for quick prototyping and plan to port it back to C later. My end goal is to incorporate it into uclogic-probe, which currently only does the older init procedures of reading string descriptors.

Also, my Star06 code is [here](https://github.com/nic3-14159/digimend-kernel-drivers/tree/Star06_new_init).

kurikaesu:
Great, having a look. From what I remember in your comments, the frame buttons are currently not working properly. Is it still the case?

nic3-14159:
No, everything is working. When I was talking about the frame buttons, I was referring to the Star 06 support from someone else that is already merged into DIGImend, I guess I was a bit ambiguous in my wording. My fork replaces the current support.

nic3-14159:
Hey @kurikaesu, could you try my new key packet script with your tablets and see if it generates expected results? Instructions are provided on the [project page](https://github.com/nic3-14159/uclogic-new-init). (I renamed the project, and the old link I sent before redirects to the new name)

kurikaesu:
Tested it on my laptop that doesn't have digimend drivers installed using my XP-Pen Deco Pro SM:

```
aren@aren-ThinkPad-T480s:~/Documents/Workspaces/uclogic-new-init$ sudo ./uclogic_new_init.py 28bd:0909
Trying init packet 0x02B00200000000000000:
Received 02B00201000000000000
Trying init packet 0x02B00400000000000000:
Received 02B00401000000000000
aren@aren-ThinkPad-T480s:~/Documents/Workspaces/uclogic-new-init$ 
```

nic3-14159:
> Tested it on my laptop that doesn't have digimend drivers installed using my XP-Pen Deco Pro SM:
> 
> ```
> aren@aren-ThinkPad-T480s:~/Documents/Workspaces/uclogic-new-init$ sudo ./uclogic_new_init.py 28bd:0909
> Trying init packet 0x02B00200000000000000:
> Received 02B00201000000000000
> Trying init packet 0x02B00400000000000000:
> Received 02B00401000000000000
> aren@aren-ThinkPad-T480s:~/Documents/Workspaces/uclogic-new-init$ 
> ```

Hmm, it should show messages indicating if a packet is invalid or correct. Just checking, you did hold the pen to the drawing surface the entire time, right? No messages could either mean that the tablet did not send any reports (which would happen if the pen was not held to the surface), or there is an issue in my code. The reason for this is that I've observed (on my tablet) that invalid keys will enable events on interface 2, but don't disable events being sent from interface 0 or 1. By holding the pen to the tablet, the program can read pen events from all interfaces and check that events are only being sent over the proper interface (usually interface 2)

kurikaesu:
Ah, that would be my fault for skimming through the documentation.
Here it is with me tapping the pen onto the surface:

```
Trying init packet 0x02B00200000000000000:
Received 02B00201000000000000
Invalid key!

Trying init packet 0x02B00400000000000000:
Received 02B00401000000000202
Key is 0x02B00400000000000000, interface 2, endpoint 0x03
```

For others like me who skim through documentation, it may be a good idea to have the tool prompt the user to tap the tablet.

nic3-14159:
>For others like me who skim through documentation, it may be a good idea to have the tool prompt the user to tap the tablet.

Done, committed, and pushed.

jbigler:
Thanks for all the hard work! I've installed driver from your branch and everything seems to be working well on my 22R Pro with the exception of the right dial.

I guess at this point my biggest peeve with the tablet is the gap in the tilt detection at 90° increments. When holding the stylus at 90/180/270/360° it can't seem decide which side to be on and makes a rather big jump of several degrees, creating ugly gaps in the drawing. I've found a way to deal with it by adjusting the tilt curves in my brushes to avoid those angles, but it is annoying. It seems to have the same problems on Windows, however, so it could just be lack of sensitivity in the hardware (which is quite disappointing if the case.)

kurikaesu:
I have seen that issue when testing tilt a little but I don't know if the actual hardware is reporting erratic values.
When I have time I'll do a more thorough deep dive on the tilt of the 22R.

kurikaesu:
@jbigler I did a quick tilt test on my 22R and recorded it here:
https://vimeo.com/manage/videos/591372066

As far as I can tell it looks like the tilt is being reported correctly and without erratic issues.

You can test identically to how I did by doing:
`libinput list`

Finding out which /dev/input/eventXX your PenDisplay Pad is under
then do a:
`sudo libinput record 2>&1 | grep -i tilt`

When you enter your password and hit enter, it will just sit there waiting for input
Type the number after /dev/input/event above ( for example, mine is `/dev/input/event19` so I would type in `19` ) and hit enter then tap & angle the pen around to see what it says.

jbigler:
@kurikaesu Thank you for testing that. So I followed your instructions and came up with some troubling results. I'm not sure where the problem lies at this point. When I tested the tilt as you instructed on a blank desktop, I initially got results as you showed. However, when I tested the tilt while using an application (I tested Krita and Inkscape), the ABS_TILT_Y tilt numbers made large jumps as they reached 0. Often from -7 up to +6 or similar, skipping up to 10 increments or more. When I ran OBS to try recording the results it jumped even more, from -11 to 10. Could it be NVIDIA?

But it isn't the whole range. As I test it more carefully the ABS_TILT_X increments quite smoothly, the jumps are occurring on ABS_TILT_Y on both the left and right sides.  Here are some sample results:

```
   - [1047, 508321,   3,  27,     -13] # EV_ABS / ABS_TILT_Y              -13 (+1)
    - [1047, 512321,   3,  27,     -12] # EV_ABS / ABS_TILT_Y              -12 (+1)
    - [1047, 542324,   3,  27,     -11] # EV_ABS / ABS_TILT_Y              -11 (+1)
    - [1047, 572322,   3,  27,     -10] # EV_ABS / ABS_TILT_Y              -10 (+1)
    - [1047, 590325,   3,  27,      -9] # EV_ABS / ABS_TILT_Y               -9 (+1)
    - [1047, 594320,   3,  27,      -8] # EV_ABS / ABS_TILT_Y               -8 (+1)
    - [1047, 598322,   3,  27,      -7] # EV_ABS / ABS_TILT_Y               -7 (+1)
    - [1047, 652345,   3,  27,      -6] # EV_ABS / ABS_TILT_Y               -6 (+1)
    - [1047, 682323,   3,  27,      -5] # EV_ABS / ABS_TILT_Y               -5 (+1)
    - [1047, 712323,   3,  27,       5] # EV_ABS / ABS_TILT_Y                5 (+10)
    - [1047, 730324,   3,  27,       6] # EV_ABS / ABS_TILT_Y                6 (+1)
    - [1047, 762324,   3,  27,       7] # EV_ABS / ABS_TILT_Y                7 (+1)
    - [1047, 818325,   3,  27,       8] # EV_ABS / ABS_TILT_Y                8 (+1)
    - [1047, 822325,   3,  27,       9] # EV_ABS / ABS_TILT_Y                9 (+1)
    - [1047, 864346,   3,  27,      10] # EV_ABS / ABS_TILT_Y               10 (+1)
    - [1047, 924327,   3,  27,      11] # EV_ABS / ABS_TILT_Y               11 (+1)
    - [1048,  76328,   3,  27,      12] # EV_ABS / ABS_TILT_Y               12 (+1)
    - [1048, 488333,   3,  27,      13] # EV_ABS / ABS_TILT_Y               13 (+1)
```

And here the jumps increased even more:

 ```
   - [965, 691409,   3,  27,     -18] # EV_ABS / ABS_TILT_Y              -18 (+1)
    - [965, 695409,   3,  27,     -17] # EV_ABS / ABS_TILT_Y              -17 (+1)
    - [965, 759601,   3,  27,     -16] # EV_ABS / ABS_TILT_Y              -16 (+1)
    - [965, 797596,   3,  27,     -15] # EV_ABS / ABS_TILT_Y              -15 (+1)
    - [965, 801601,   3,  27,     -14] # EV_ABS / ABS_TILT_Y              -14 (+1)
    - [965, 879598,   3,  27,     -13] # EV_ABS / ABS_TILT_Y              -13 (+1)
    - [965, 883411,   3,  27,     -12] # EV_ABS / ABS_TILT_Y              -12 (+1)
    - [966, 141606,   3,  27,     -11] # EV_ABS / ABS_TILT_Y              -11 (+1)
    - [966, 205606,   3,  27,      10] # EV_ABS / ABS_TILT_Y               10 (+21)
    - [966, 579419,   3,  27,      11] # EV_ABS / ABS_TILT_Y               11 (+1)
    - [966, 631419,   3,  27,      12] # EV_ABS / ABS_TILT_Y               12 (+1)
    - [966, 733421,   3,  27,      13] # EV_ABS / ABS_TILT_Y               13 (+1)
    - [966, 745421,   3,  27,      14] # EV_ABS / ABS_TILT_Y               14 (+1)
    - [966, 873422,   3,  27,      15] # EV_ABS / ABS_TILT_Y               15 (+1)
    - [966, 893608,   3,  27,      16] # EV_ABS / ABS_TILT_Y               16 (+1)
```

Changing to another USB port had no affect. I may have to try configuring another system to do more tests.
Currently, I'm running i3 on Arch Linux with an AMD Ryzen 7 3700X CPU and an NVIDIA GeForce GTX 1660 SUPER graphics card (proprietary NVIDIA drivers.)

**Comment by kurikaesu** [comment]
Interesting, I re-performed the test just oscillating my stylus on the Y axis but keeping X axis rotation 0.
In Krita and on a blank desktop, I also see bursts of tilt changes however on my side it does not go from -11 to 10. Instead it will burst -9. -8, -4, 0, 1, 4, 7, 9, 11, 15, 18, 20 and so on very very quickly. 

From my observations it seems like whatever is in the pen that reports tilt gets too far away from the digitizer electric/magnetic field when the stylus is close to perpendicular to the screen. It will then at least for my case blast all of the stored angle changes at once when it gets back into range.

I don't think this is something that we can deal with on a driver level. Looks like a design defect on XP-Pen's side.

**Comment by kurikaesu** [comment]
After testing my changes the past few days there is something different with the Art Deco Pro. While the report descriptors appear to be correct and all functions work, the tablet itself does not immediately start working when plugged in. I have to use the dumper tool to sort of activate the tablet after which it will work fine until I disconnect and plug it in again.

I'm guessing there is a piece of the handshake puzzle missing in the digimend driver that the dumper does

**Comment by kanashimia** [comment]
Deco pro M has different product id: `0904` 
See https://github.com/DIGImend/digimend-kernel-drivers/issues/366 for more info dumps.

**Comment by kanashimia** [comment]
> After testing my changes the past few days there is something different with the Art Deco Pro. While the report descriptors appear to be correct and all functions work, the tablet itself does not immediately start working when plugged in. I have to use the dumper tool to sort of activate the tablet after which it will work fine until I disconnect and plug it in again.
> 
> I'm guessing there is a piece of the handshake puzzle missing in the digimend driver that the dumper does

I have same problem, and just doing `sudo modprobe -r hid-uclogic && sudo modprobe hid-uclogic` helps.
What is the dumper tool you are talking about?

**Comment by kurikaesu** [comment]
Thanks for reporting the ID. I've added it into my PR as one of the supported devices.

When using libinput it is normal for it to be unbound since the pads buttons are just "button_0", "button_1" etc.
It allows you to fully bind the pad to do whatever. This is in contrast to using the default bindings that are available but restrict you on what you can change the bindings to. For example, if a button produces a keypress of "B" then it can only be substituted by another single keypress. Can't be changed into something like "shift+space" as an example.

The "button_0", "button_1"... unbound libinput gives you full flexibility at the cost of having no default bindings as I haven't provided any.

Dumper tool I am talking about is usbhid-dump: https://github.com/DIGImend/usbhid-dump
Used like `sudo usbhid-dump --entity=stream --model=28bd`

Ideally the driver should not have to be unloaded in order for the tablet to work when plugged in (this is what modprobe -r does)

**Comment by kurikaesu** [comment]
Latest commit fixes the deco pro small (and probably medium) on-connect issues. The tablet for whatever reason sends through a blank message on Report ID 1 and report ID 6 before switching over to using the unified interface. This causes the driver to block receiving new messages until the interface is reset by either using a tool to forcefully detach and claim the device and release it or other methods.

Implementing the report descriptors for interface 0 which receives the report id 1 and 6 fixes the connection issues. This means that the tablet should start working on plug in without having to do any fiddling around (other than of course making sure to bind all of the buttons and wheel).

There is something that has changed recently and I don't know why but the touch-dial on the Deco Pro is now reporting as a mouse input on interface 0 (even if we use the key to get all the stylus movements, buttons and dial reporting on interface 2). Very odd! I tested this behavior before my most recent commit so the changes I added didn't introduce this weird behavior.

**Comment by kurikaesu** [comment]
For those who come here and have an Artist 22R Pro, consider testing out my userland drivers here: https://github.com/kurikaesu/xp-pen-userland

It comes with a configuration gui here: https://github.com/kurikaesu/xp-pen-userland-config-util

I'll add support for all of the devices listed in this PR in my userland drivers as time goes on.

**Comment by alfazaz** [comment]
Thanks in advance for your work for the artist 12 pro which I will be able to test when it will be released.

**Comment by kurikaesu** [comment]
I have just added support for the Artist 12 Pro and the 24 Pro but I can't test them since I don't own either product.
If you do test them and find issues, create an issue in my corresponding repository. Don't post issues here.

**Comment by kanashimia** [comment]
So what i have found out:

When you hover with pen then both wheels don't work, it is like if touch arbitration is in place, really annoying.
There is no such problem when using wacom driver.

Wheels don't work while using wacom driver, but when you press pads many times very fast - physical dial starts working, while touch dial doesn't. Not sure what is going.

By default udev assigns `ID_INPUT_KEY=1` to pads, but if you modify hwdb like this:
```
id-input:modalias:input:b0003v28BDp0904e0100-e0,1,2,4*r6*
  ID_INPUT_TABLET=1
  ID_INPUT_TABLET_PAD=1
  ID_INPUT_KEY=0
 ```
(supposed to match only *Pad) then pads start emitting events, and wheels stop working.
Changing `ID_INPUT*` stuff doesn't seem to do anything while using wacom driver.

kurikaesu:
The Deco Pro is pretty special in how it functions behind the scenes. It appears there is an on-device bit that can be set by the official driver. I haven't captured the bit yet, just tested the effects of switching it around.

The process I went through was:
Plug Deco Pro into Windows computer => Change touch to <X> and hit OK => Unplug from Windows computer and plug into Linux computer => Execute `sudo usbhid-dump --entity=stream --model=28bd` and note the stream output.

Just for reference, interface 1 is 0x00.

When <X> is set to "mouse", outputs come out of interface 2 (0x01) and produce X, Y coordinates:
```
001:015:000:STREAM             1631326995.444260
 01 00 FB FF FF FF 00 00
```

When <X> is set to "roller", then it still outputs from interface 2 (0x01) but emulates the scroll wheel:
```
001:016:000:STREAM             1631327086.714873
 01 00 00 00 00 00 01 00

001:016:000:STREAM             1631327086.898879
 01 00 00 00 00 00 FF 00
```

When <X> is set to "customized" then it will output from interface 3 (0x02) and share the same bitmap field as the dial:
```
001:017:002:STREAM             1631327213.877488
 02 F0 00 00 00 00 00 04 00 00

001:017:002:STREAM             1631327214.047540
 02 F0 00 00 00 00 00 08 00 00
```

If you haven't used the tablet with the official drivers yet and changed the default behavior then the tablet will probably be in mouse mode. Since mouse mode and roller mode share the same interface, we would have to have bytes 3 to 7 all mapped in the resource descriptor.

If you are using the code before https://github.com/DIGImend/digimend-kernel-drivers/pull/557/commits/beb79ff3d92ba500740e5657ca29b9296977f8e1 then whenever you use the touch pad it will likely be sending a message through interface 2 which the driver up until the above commit doesn't read. Since it doesn't parse it the whole device "hangs" while it waits for something to parse the message.

The above commit will interpret the message as mouse movements but that isn't mapped to anything and its unlikely the wacom driver will know what to do with it either.

I'm not providing a fix for this in this DIGImend PR, just writing out research notes and explanations.

kurikaesu:
Just finished adding support for the Deco Pro SM and MD (MD is unconfirmed, someone please test).
I don't send in the necessary message to change the touch-pad to "Customized" which means it will support the tablet in whatever configuration it is set at, be it "mouse", "roller" or "customized".

Configuration GUI also supports setting up the bindings but again, if the touch-pad is not set to customized then the touch pad binding will not be in effect.

kanashimia:
> [..] but if you modify hwdb like this:
> [..] then pads start emitting events, and wheels stop working.

I figured it out, it is a problem with libinput, it doesn't handle EV_REL events for tablet pad, you can see it when you run `libinput debug-events`

PR #577, opened 2021-11-06
XP-PEN Deco 03 Support

zetagabriels:
Basic support for the Deco 03.

### What works

Pen pressure and all three stylus buttons (`xsetwacom`)

### What does not

Pad's frame buttons and frame wheel. `xsetwacom` can see the pad using `list devices` but changing the button values has no effect


kurikaesu:
Does this particular model have a unified interface mode that can be set?

You can use this tool to see if it has a initializable unified interface mode: https://github.com/nic3-14159/uclogic-new-init

If it does then let me know and I'll also add support for the tablet to my driver here: https://github.com/kurikaesu/userspace-tablet-driver-daemon

zetagabriels:
@kurikaesu Here is the output. This was done while the DIGImend driver was installed, and after running it, the tablet needed to be reconnected.
If you need any more info, let me know


[init.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/7502638/init.txt)

PR #594, opened 2022-02-03
Makefile improvements

albfan:
This is a series of fixes and improvements for makefile:

- detect dkms module and version: for me digimend/11 is not correctly splitted
- avoid set -e
- uninstall module on install (to reset it effectively)
- silent execution of dkms (using @)
- automatize dkms uninstall, install, and set debug mode (new target dkms_reinstall)

Let me know if any of this makes sense to cherry-pick or create new isolate PR


Makefile
```diff
@@ -126,5 +126,10 @@ dkms_install: dkms_modules_install depmod_conf_install dracut_conf_install udev_
 
 dkms_uninstall: tools_uninstall xorg_conf_uninstall udev_rules_uninstall dracut_conf_uninstall depmod_conf_uninstall dkms_modules_uninstall
 
+enable_debug:
+	./digimend-debug 1
+
+dkms_reinstall: dkms_uninstall dkms_install enable_debug
```
spbnick, line 132: Hmm, I assume this would be useful for a development loop, to quickly test your changes. However, enabling debug shouldn't be a part of a general-purpose target. After removing it, its use for debugging becomes superficial, and you can just instead have a compound command in your history to do the same:

    make dkms_uninstall dkms_install && digimend-debug 1

All-in-all, I think it's more trouble than it's worth, so could you please drop this patch?

Makefile
```diff
@@ -108,7 +108,7 @@ dkms_modules_uninstall: dkms_check
 	dkms status $(DKMS_MODULES_NAME) | \
 	    while IFS=':' read -r modules status; do \
 	        echo "$$modules" | { \
-	            IFS=', ' read -r modules_name modules_version \
```
spbnick, line 113: Thanks! I'm curious, though, where does that change come from? Has `dkms status` output format changed?

Makefile
```diff
@@ -102,6 +102,7 @@ dkms_modules_install: dkms_check
 	dkms add .
 	dkms build $(DKMS_MODULES)
 	dkms install $(DKMS_MODULES)
+	@! rmmod hid_uclogic
```
spbnick, line 105: I'm hesitant about loading or unloading modules as part of (un)installation, since we don't really know which version of the module is loaded at that moment. At the very least this should be a part of an uninstall target, but I'd prefer, if you dropped this patch.

spbnick (changes requested):
Thank you, Alberto!
There are quite a few useful changes, but I've left a few inline comments asking to drop some things, if possible.

albfan:
I will rebase on top of #629 and address your suggestions.

Thanks for taking a look

PR #598, opened 2022-02-06
Changes from upstream

JoseExposito:
Patches present upstream (hid for-5.20/uclogic) but not in DIGImend, including the KUnit test. Let me know if you want me to remove those, as in DIGImend are not easy to run.

I included the patch https://github.com/JoseExposito/linux/commit/f7271b2a697d2549db7aa60e0a8e0afeb852faa5 even though it removes a quirk from `drivers/hid/hid-quirks.c`. @spbnick  I'm not sure if this could break the tablet in old kernels, let me know your opinion. It could be a good idea to revert it just in case.

This PR includes adds support for the XP-Pen Deco L. The device basic information is present in https://github.com/DIGImend/tablets/pull/52.

---

There are 3 patches that'd need to be backported that I didn't apply because could break compilation in old kernels:

Use fallthrough: https://github.com/JoseExposito/linux/commit/df561f6688fef775baa341a0f5d960becd248b11

hid_is_usb: https://github.com/JoseExposito/linux/commit/f83baa0cb6cfc92ebaf7f9d3a99d7e34f2e77a8a

hid_is_usb (Not sure if we could apply this one): https://github.com/JoseExposito/linux/commit/93020953d0fa7035fd036ad87a47ae2b7aa4ae33


albfan:
So what is the point? Is this repo maintained?

I report some changes here, but looks all work is happening in kernel repos, isn't it?

nic3-14159:
> So what is the point? Is this repo maintained?
> 
> I report some changes here, but looks all work is happening in kernel repos, isn't it?

This repo isn't officially unmaintained, but it hasn't seen much activity from the official maintainer for a while.

This repo is essentially upstream for the drivers in the kernel repo. The intent is that any changes here eventually make their way into the kernel. However, there's nothing stopping people from just directly submitting code to the kernel, so once in a while it's good to pull those changes back here.

Looking at the kernel logs, most of the major changes came from here. more recent commits directly to the kernel are relatively small, and generally don't seem related to the issues posted to this repo. Even if patches don't get merged in DIGImend for a while, it's likely faster for users to just build from whatever pull requests people make here instead of waiting for the change to make its way into the kernel of their distro.

albfan:
Yes, that's what I though, let's see. I will try both ways

What's the mailing list to report changes to this repo in kernel? I'm reading https://www.kernel.org/doc/html/latest/process/maintainers.html#tips-for-patch-submitters looking for a review. Any suggestion?

JoseExposito:
> So what is the point? Is this repo maintained?

I'm not sure about the maintenance status of the repo, but since I need to send to the kernel some of the patches present in this repo for my tablet, I decided to contribute in the other direction too with the intention of having the same code in the kernel and DIGImend.

hid-uclogic-params.c
```diff
@@ -1008,6 +1008,197 @@ static int uclogic_params_huion_init(struct uclogic_params *params,
...
+		hid_err(hdev, "failed retrieving pen and frame parameters: %d\n", rc);
+		uclogic_params_init_invalid(&p);
+		goto output;
+	}
+
+	desc_params[UCLOGIC_RDESC_PEN_PH_ID_X_LM] =
+		get_unaligned_le16(str_desc + 2);
+	desc_params[UCLOGIC_RDESC_PEN_PH_ID_Y_LM] =
+		get_unaligned_le16(str_desc + 4);
+	desc_params[UCLOGIC_RDESC_FRAME_PH_ID_UM] = str_desc[6];
```
spbnick, line 1133: Looking at different tablets from Huion, we saw no correlation between the number of buttons on the frame and any data in the configuration response. That's one of the reasons I didn't go with exposing that in the descriptor. How many Ugee/XP-Pen tablets have you seen so far, that correctly report their frame button number here?

spbnick (approved):
Thank you very much for getting this done, José!
This looks good to me, will merge in a second.

However, I have one non-urgent question inline, and also would like to know exactly how would one run the kunit tests out-of-tree, in DIGImend? And do you think you could add at least some instructions for that to README.md?

spbnick:
Oh, and is there anything else you'd like merged before I make the release, José?

hid-uclogic-params.c
JoseExposito: So far, I have seen reports from: XP-Pen Deco Mini 4, XP-Pen Deco Mini 7, XP-Pen Deco L, XP-Pen Deco Pro S, XP-Pen Deco Pro SW and Parblo A610 PRO (UGEE as well as XP-Pen *).

In all cases the number of buttons is properly reported. Thankfully, these tablets have different VIP:PID, so we could quirk a specific device is required.

JoseExposito:
Hi @spbnick,

Thanks for merging this :)

> However, I have one non-urgent question inline, and also would like to know exactly how would one run the kunit tests out-of-tree, in DIGImend? And do you think you could add at least some instructions for that to README.md?

As far as I know, it is not possible to run KUnit tests for out-of-tree modules... At least without [doing something complicated](http://www.philipmolloy.com/kunit.html).

I decided to leave the KUnit test files for 2 reasons:

The first one is to make them more discoverable for DIGIment contributors, they might want to add new tests for their features and run them in the kernel.

The second one is to be able to backport patches easily from/to the kernel, without having to remove the test files from the patch.

However, feel free to remove the test files if you think that it doesn't make sense to have them in DIGImend (or I can open a PR removing them if you are busy).

> Oh, and is there anything else you'd like merged before I make the release, José?

Not for the moment. There are some patches awaiting for review in the mailing list, but I'd prefer to wait until they are accepted before merging them here.

By the way, I didn't CCed you because I didn't want to bother you, but you might be interested in this regression in `hid-input.c` affecting HUION devices:
https://lore.kernel.org/linux-input/20220727162755.GA3839@elementary/T/#e2063d77e49e578b75896f06099977e4edb0a419d

Thanks,
Jose

PR #605, opened 2022-03-09
We add support for the newer Xppen tablet mini7.

MikuChan03:
The thing is that the tablet reports the buttons as keyboard events.
This can only be circumvented by using the vendor defined
usage and usage page, for which the events are reported in a
unified format. To activate this interface, we need to send a
a certain key on a certain endpoint, shown in:

https://github.com/nic3-14159/uclogic-new-init

Still this interface is opaque in that the report descritpor only
shows the length, not what the individual bits do.
Therefore we need to add two more devices, one for the pen and
one for the buttons. The first interface doesn't go away but
doesn't report anything either, since it's reports are translated
into the other two devices. I have called this original device
"VendorDevice".

Once all this is done, native support with the wacom driver
is achieved. Though since the wacom driver doesn't recognize
the IDs yet, we are in need of X config file like:

Section "InputClass"
  Identifier "xppen wacom all"
  MatchUSBID "28bd:0928"
  Driver "wacom"
EndSection

That's all.


nic3-14159:
Since a62702fcdb026ed0cc7e3ad8911cdb7730c61080 has already been merged, this PR (mainly hid-uclogic-params.c) should probably be adapted to use code from that commit, as that commit already implements the new initialization procedure for these newer XP-Pen/Ugee tablets.

PR #627, opened 2022-08-03
To fix dkms remove issue.

uhuntu:


spbnick:
Thank you for the PR, @uhuntu. However, could you please describe what is the problem you're trying to solve, and why your fix works? Thank you!

nic3-14159:
Just a note, I think this accomplishes the same thing as 60f62f50834d6e52654b84af1592568f5ec15873 in #594

spbnick:
Ah, thank you, @nic3-14159. I'll take a look at that.

spbnick:
Fixed in #629. Thank you!

PR #682, opened 2024-04-14
Support HUION devices with up to 20 buttons

JoseExposito:
The HID descriptor for HUION tablets was limited to 13 buttons. However, there are devices with more buttons in their frames.

So far, the device with more buttons released by HUION is the Huion Kamvas Pro 24 (GT-240, QHD) [1], with 20 buttons.

Tweak the HID descriptor to support it.

Link: https://gitlab.freedesktop.org/libinput/libinput/-/issues/989  [1]


JoseExposito:
2 different users tested it and the result was successful:
https://gitlab.freedesktop.org/libinput/libinput/-/issues/989

Merging!

PR #683, opened 2024-04-19
Use Rx and Ry for touch strips

JoseExposito:
Currently, HUION devices use ABS_WHEEL as the usage for touch strips.

There are 2 main issues with this approach:

The first one is that the descriptor for touch rings (uclogic_rdesc_v2_frame_touch_ring_arr) also uses ABS_WHEEL. From user-space it is impossible to know which device sends the events.

The second one is that Wacom uses ABS_RX/ABS_RY to notify events from touch strips and user-space was designed to handle those axes.

Change the usage of touch strips to Rx/Ry to fix both issues.

Link: https://gitlab.freedesktop.org/libinput/libinput/-/issues/989

### DIFF
--- a/hid-uclogic-rdesc.c
+++ b/hid-uclogic-rdesc.c
@@ -789,7 +789,8 @@ const __u8 uclogic_rdesc_v2_frame_touch_strip_arr[] = {
 	0x95, 0x01,         /*          Report Count (1),           */
 	0x81, 0x02,         /*          Input (Variable),           */
 	0x05, 0x01,         /*          Usage Page (Desktop),       */
-	0x09, 0x38,         /*          Usage (Wheel),              */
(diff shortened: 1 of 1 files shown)


AesaraB:
[libinput-record-yes-patchRxRy-strip.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/15076307/libinput-record-yes-patchRxRy-strip.txt)
[libinput-record-yes-patchRxRy-pad.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/15076310/libinput-record-yes-patchRxRy-pad.txt)

JoseExposito:
Thanks a lot for your testing and for the recordings @AesaraB. Merging!

I'll try to create a DIGImend release next week including the changes to fully support your tablet.

PR #686, opened 2024-05-04
Do not attempt to build modules for kernels without CONFIG_USB_HID

kenhys:
It was forwarded patch from
https://salsa.debian.org/debian/digimend-dkms/-/merge_requests/1

This patch was originally written by Andreas Beckmann <anbe@debian.org>


JoseExposito (approved):
Thanks for backporting this fix, merging!

PR #687, opened 2024-05-04
Delegate placement of udev files to pkg-config data

kenhys:
delegate the exact placement of the udev files to udev.pc (using pkg-config)

This patch was originally written by Chris Hofstaedtler <zeha@debian.org>

FYI:
https://bugs.debian.org/cgi-bin/bugreport.cgi?att=1;bug=1057748;filename=digimend-dkms_11-3-udevdir.patch;msg=5

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1057748


JoseExposito (approved):
Thanks! Tested on Fedora and it works as well.

PR #688, opened 2024-05-06
Add support for XP-Pen Deco M

Pixelthegreat:
Since the Deco L was already on the list of supported devices, I figured the Deco M wouldn't be hard to add, considering it is just the Deco L in a different size variant.


hid-ids.h
```diff
@@ -63,6 +63,7 @@
 #define USB_DEVICE_ID_UGEE_XPPEN_TABLET_G640	0x0094
 #define USB_DEVICE_ID_UGEE_XPPEN_TABLET_DECO01	0x0042
 #define USB_DEVICE_ID_UGEE_XPPEN_TABLET_DECO01_V2	0x0905
+#define USB_DEVICE_ID_UGEE_XPPEN_TABLET_DECO_M	0x0936
```
JoseExposito, line 66: Place it after the Deco L ID, please. We try to keep the list sorted alphabetically

hid-uclogic-params.c
```diff
@@ -1279,10 +1279,10 @@ static bool uclogic_params_ugee_v2_has_battery(struct hid_device *hdev)
 	 * version.
 	 */
 	if (hdev->vendor == USB_VENDOR_ID_UGEE &&
-	    hdev->product == USB_DEVICE_ID_UGEE_XPPEN_TABLET_DECO_L) {
+	    hdev->product == USB_DEVICE_ID_UGEE_XPPEN_TABLET_DECO_L || hdev->product == USB_DEVICE_ID_UGEE_XPPEN_TABLET_DECO_M) {
```
JoseExposito, line 1282: Do you have a Deco MW? This makes sense, but sometimes vendors surprise you and change things.
If this is untested, I'd prefer to not change it. Someone with a Deco MW can always fix and test this in the future.

JoseExposito (changes requested):
Hi @Pixelthegreat 

Thanks for your patch! Overall, it looks very very good.

I recently got permission to merge code and I didn't have time yet to include basic commit message validation. Could you change your commit message to look like:
https://github.com/DIGImend/digimend-kernel-drivers/commit/72ba179db1142cfe224e7195c884e28f53a10dc7

The important bits are "HID: uclogic:" and the "Signed-off-by:" with your name an email. These are requirements to add your code to the kernel.

hid-uclogic-params.c
Pixelthegreat: I do indeed have an MW, however, I am not certain of how to test if this device works with this quirk, so I will revert it.

Pixelthegreat:
Due to some confusion with merging the repository, I'm going to reset my fork and create a new pull request. I probably could've avoided it if I had read this a week ago, but I am not super familiar with operating pull requests and stuff, and I would prefer to redo this clean with the necessary changes.

hid-uclogic-params.c
JoseExposito: If you see the battery level and status (charging/discharging) when you connect your device, it means that no changes are required.

However, if this method returns false for your MW, you'll need to add a condition similar to this one.

In GNOME, the battery is displayed in System settings > Power. Other desktop environments display it in other places.

PR #690, opened 2024-05-13
HID: uclogic: Add Huion Inspiroy 2 S (H641P)

whot:
Unlike the M and L version added in commit 2683e80de827 ("HID: uclogic: Add support for Huion Inspiroy 2 M and L") the small version does not have the group buttons. This patch only adds the ID, it doesn't try to remove that device.

cc @tequeter, @JoseExposito
Related to #679

---

Hid report descriptors:
- [hidraw1.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/15288745/hidraw1.txt) - vendor page only
- [hidraw2.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/15288746/hidraw2.txt) the pen device
- [hidraw3.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/15288747/hidraw3.txt) the keyboard and consumer device collections

I admit from the quick hacking I did I'm not 100% how the group mode buttons device is supposed to work but my gut feeling (having implemented this for wacom devices 10y ago) is that having a separate kernel device here is not ideal.


JoseExposito (approved):
Thanks! CC me if you upstream this patch, please. Otherwise, I'll upstream Inspirioy 2 (S) support in a few weeks after we have some testing.

whot:
will do, I'm still hoping for some feedback on the button mode discussino from #679, might have to change the patch pending that.

PR #691, opened 2024-05-14
HID: uclogic: Added support for XP-Pen Deco M

Pixelthegreat:
This is the same as the Deco L, with the only difference (as far as I know) being the physical size.

I made a new pull request for this due to some issues with the prior one that I didn't know how to work around.


JoseExposito (approved):
Thanks a lot!

For the next time, you can edit the commit message and the changes in your last commit with `git commit --amend`. You'll need to force push your changes (`git push -f`). Make sure to avoid doing it in shared branches (main/master or similar) but for pull request it is a common practice.

Pixelthegreat:
Thanks!

PR #693, opened 2024-05-19
Draft: Use HID_QUIRK_MULTI_INPUT 

JoseExposito:
Userspace requires the pad buttons, touch-rings, touch-strips, dials and similar devices to be in the same event node.
    
Use HID_QUIRK_MULTI_INPUT to expose only 2 event nodes: one for the pen and another one for the pad devices.

---------------------------------

You'll notice that, in regular HUION devices, there are 3 interfaces:

- Keyboard: I don't know if this one is used by any device, but (and least in my H640P) it is garbage from the firmware
- Keypad: Pad buttons, touch-rings, touch-strips, dials, etc
- Unnamed: The pen

On XP-Pen:

- Keypad: Pad buttons, touch-rings, touch-strips, dials, etc
- Unnamed 1: The pen
- Unnamed 2: Mouse. It is the circular area inside the dial in devices like the [Deco Pro SW](https://www.storexppen.es/buy/deco-pro-sw.html). (Which by the way stopped working in working with "recent" libinput).

Mark as draft because:

- The naming of the interfaces is broken because `uclogic_input_configured` doesn't receive the hid-report anymore. I need to investigate if it is possible to fix it, but I run out of time.
  As a side effect of the naming changes, libwacom matches will break and custon user config might break as well.

- `EV_MSC` reports are not disabled anymore for touch-rings.  I can't test it but, if there is any kind of support in userspace, this change could break it:
   https://github.com/DIGImend/digimend-kernel-drivers/blob/master/hid-uclogic-core.c#L122-L127

I'm not sure about merging this. Unless the benefit (i.e., dials, touch-rings, etc work auto-magically) is grater than the possible issues caused, we should handle this PR carefully.

cc @spbnick @tequeter @whot

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ website][website].
 Installing
 ----------
 
-Kernel v3.5 or newer is required.
+Kernel v4.18 or newer is required.
 
 Download appropriate files for one of the releases from the [releases
 page][releases]. The "Download ZIP" link on the right of the GitHub page leads
--- a/hid-uclogic-core.c
+++ b/hid-uclogic-core.c
@@ -83,15 +83,8 @@ static int uclogic_input_mapping(struct hid_device *hdev,
 	return 0;
 }
 
-#if KERNEL_VERSION(4, 4, 0) > LINUX_VERSION_CODE
-#define RETURN_SUCCESS return
-static void uclogic_input_configured(struct hid_device *hdev,
-		struct hid_input *hi)
-#else
-#define RETURN_SUCCESS return 0
 static int uclogic_input_configured(struct hid_device *hdev,
 		struct hid_input *hi)
-#endif
 {
 	struct uclogic_drvdata *drvdata = hid_get_drvdata(hdev);
 	struct uclogic_params *params = &drvdata->params;
@@ -102,7 +95,7 @@ static int uclogic_input_configured(struct hid_device *hdev,
 
 	/* no report associated (HID_QUIRK_MULTI_INPUT not set) */
 	if (!hi->report)
-		RETURN_SUCCESS;
+		return 0;
 
 	/*
 	 * If this is the input corresponding to the pen report
@@ -158,10 +151,8 @@ static int uclogic_input_configured(struct hid_device *hdev,
 		hi->input->name = devm_kasprintf(&hdev->dev, GFP_KERNEL,
 						 "%s %s", hdev->name, suffix);
 
-	RETURN_SUCCESS;
+	return 0;
 }
-#undef RETURN_SUCCESS
-
 
 static int uclogic_probe(struct hid_device *hdev,
 		const struct hid_device_id *id)
--- a/hid-uclogic-core.c
+++ b/hid-uclogic-core.c
@@ -162,10 +162,10 @@ static int uclogic_probe(struct hid_device *hdev,
 	bool params_initialized = false;
 
 	/*
-	 * libinput requires the pad interface to be on a different node
-	 * than the pen, so use QUIRK_MULTI_INPUT for all tablets.
+	 * libinput requires the pen interface to be on a different node
+	 * than the pad, dials, touch strips, etc.
 	 */
-	hdev->quirks |= HID_QUIRK_MULTI_INPUT;
+	hdev->quirks |= HID_QUIRK_INPUT_PER_APP;
 	hdev->quirks |= HID_QUIRK_HIDINPUT_FORCE;
 #ifdef HID_QUIRK_NO_EMPTY_INPUT
 	hdev->quirks |= HID_QUIRK_NO_EMPTY_INPUT;


whot:
`s/HID_QUIRK_MULTI_INPUT/HID_QUIRK_INPUT_PER_APP/` in the description/commit message?

But yeah, I agree, enabling this across the board would not be a good idea, it will cause regressions that we can only find by testing every single device. Enabling it for new devices - different matter. Though right now I'm still thinking that a bpf-based driver might be the easier way to go.

> It is the circular area inside the dial in devices like the [Deco Pro SW](https://www.storexppen.es/buy/deco-pro-sw.html). (Which by the way stopped working in working with "recent" libinput).

oh? that'd be a bug - isn't that a REL_WHEEL? ow what does it send?

> The naming of the interfaces is broken

from my quick investigation, the `Pad` suffix requires more driver support too, we can't rely on hid-generic/core for that.

> EV_MSC reports are not disabled anymore for touch-rings. 

after having a (sad) look at the wacom X driver: if `EV_MSC` is set, then the ring defaults to 0..71 as opposed to the evdev min/max range. let's assume historical reasons but be sad anyway.


I think one of the discussions to be had is: digimend is an old project and originated against a quite different environment than is now. So while we cannot break compat for every tablet already supported, we should ask if the current approach is the best for supporting new tablets. e.g. for the inspiroy - the one I have access to - the support was a fair bit of cookie-cutter additions but that also leaves us with a device structure not well supported in userspace. It might be worth (and note that I don't have experience here beyond the last few days, so I might be off) revisiting what a slimmer/simpler driver support could look like in the future (whether that's bpf or just sliming uclogic down).

JoseExposito:
> s/HID_QUIRK_MULTI_INPUT/HID_QUIRK_INPUT_PER_APP/ in the description/commit message?

Indeed, thanks.

> But yeah, I agree, enabling this across the board would not be a good idea, it will cause regressions that we can only find by testing every single device. Enabling it for new devices - different matter.

Yes, that is definitely possible thanks to firmware version or PIDs for the new devices.

I'm not sure if it'd help userspace though. If we want to support "old" devices (old == devices with VIP/PID not in the kernel right now), we'd need to support events split across nodes.

If the idea is to create a new driver for new devices and stop adding devices in UCLogic then it'd work. But sadly, that would mean that we give up on a ton of hardware out there.

> Though right now I'm still thinking that a bpf-based driver might be the easier way to go.

The only problem I see with BPF is that, as far as I know, you can't get the firmware name so, if vendors keep reusing PIDs, we might end up with a generic BPF driver.

> I think one of the discussions to be had is: digimend is an old project and originated against a quite different environment than is now. So while we cannot break compat for every tablet already supported, we should ask if the current approach is the best for supporting new tablets. e.g. for the inspiroy - the one I have access to - the support was a fair bit of cookie-cutter additions but that also leaves us with a device structure not well supported in userspace. It might be worth (and note that I don't have experience here beyond the last few days, so I might be off) revisiting what a slimmer/simpler driver support could look like in the future (whether that's bpf or just sliming uclogic down).

This makes sense in the second scenario I described: we stop working on "old" devices and focus on supporting new hardware. If that's the case then making a new driver looks like the best solution to me. It'd be way easier to add new tablets without worrying about regressions.

Also, it'd allow stuff like Bluetooth support or a good quirk system now that we are able to know which device is connected.

I have mixed feeling about starting a new driver. If we take this route, I'd like to document what userspace expects from the kernel and not just "what the Wacom driver does". Also, we'd need better tooling. We can not rely on someone connecting a bunch of tablets and calling the day.

whot:
> I'm not sure if it'd help userspace though. If we want to support "old" devices (old == devices with VIP/PID not in the kernel right now), we'd need to support events split across nodes.

yeah, eventually. but that is a large amount of work that (judging by how much outside contributions *any* of the tablet code in libinput/gnome/... gets) will take a long time. Meanwhile, we may get other tablets supported quickly.

> If the idea is to create a new driver for new devices and stop adding devices in UCLogic then it'd work. But sadly, that would mean that we give up on a ton of hardware out there.

I wonder if it's more selective than this. There will be more cheap devices that may require the uclogic and there will be others that behave better and can thus be handled better. So uclogic would become more of a special-case/compatibility driver with the better behaved tablets being handled by a simpler driver approach?

> and not just "what the Wacom driver does"

On the positive side, the xf86-input-wacom driver (just like Xorg) is in maintenance mode so we will have to care less and less over time about its accumulated quirks. libinput is a bit easier to update and test for new devices.

Having said that, updating the wacom driver to special-case the `EV_MSC` behaviour to be limited to the Wacom VID only is a possibility.

> Also, we'd need better tooling. We can not rely on someone connecting a bunch of tablets and calling the day.

On the hid level we have fairly good tooling, both with hid-recoder/hid-replay and with the hidtools kernel test suite that @bentiss is even updating to accommodate for the BPF programs we're adding. That tooling needs to be taken advantage of though.

> The only problem I see with BPF is that, as far as I know, you can't get the firmware name so, if vendors keep reusing PIDs, we might end up with a generic BPF driver.

yes-ish. with udev-hid-bpf and the way it's designed is that we trigger the BPF load through a udev rule and a hwdb match. that currently uses name + vid/pid but we could expand this to anything we can find in sysfs. I'm not sure if it'd be possible to load a driver that extracts the firmware, pushes it into a userspace visible value and then rebind based on that via the udev rules. @bentiss?

bentiss:
>> If the idea is to create a new driver for new devices and stop adding devices in UCLogic then it'd work. But sadly, that would mean that we give up on a ton of hardware out there.

No, we are not dropping hid-uclogic nor adding a new driver. We are just empowering the work done here in a much more efficient way. If we rely on hid-bpf, we don't need to update the kernel, and we don't need to recompile a kernel module for a particular device. We create the file, add it to udev-hid-bpf, and done. It's fixed for users without having to recompile anything.

There might be cases where we need a kernel update, but they are very limited, so HID-BPF will allow for a much easier user experience for end-users.

> yes-ish. with udev-hid-bpf and the way it's designed is that we trigger the BPF load through a udev rule and a hwdb match. that currently uses name + vid/pid but we could expand this to anything we can find in sysfs. I'm not sure if it'd be possible to load a driver that extracts the firmware, pushes it into a userspace visible value and then rebind based on that via the udev rules. @bentiss?

Even simpler: hid-bpf has access to all the fields of `struct hid_device`. So the `probe()` function has access to `.name[]`, `.phys[]` and `.uniq[]`. So if hid-uclogic.ko just sets those fields and be done, that allows hid-bpf to get the firmware version.

We can also have one separate hid-bpf program for every new device, and let `.probe()` fetch that firmware information already stored in `struct hid_device`. But in that case, we might as well rely on the sysfs as @whot said (but we can cross that bridge when we need it).

But honestly, for those shared device, we better stick to hid-uclogic, or have an hybrid solution: fix what the hid-uclogic.ko driver is doing in a HID-BPF fashion.

Last, but not least, I intend to add kernel tests for any hid-bpf program we ship upstream. So that means that we will have regression tests for those devices, allowing for a better "no regression" rule.

JoseExposito:
> No, we are not dropping hid-uclogic nor adding a new driver. We are just empowering the work done here in a much more efficient way.  If we rely on hid-bpf, we don't need to update the kernel, and we don't need to recompile a kernel module for a particular device. We create the file, add it to udev-hid-bpf, and done. It's fixed for users without having to recompile anything.
> 
> There might be cases where we need a kernel update, but they are very limited, so HID-BPF will allow for a much easier user experience for end-users.

I absolutely agree on how easy it is for end users to install and test a HID-BPF program compared to compiling the kernel.
However, with DIGImend, the same is possible installing a deb package or running `sudo make install`. While in the general case HID-BPF is a huge improvement, in this case is not as big of a change.

Also, DIGImend supports kernel >= v3.5 but HID-BPF won't work in certain distros. In the long run it won't matter, but at the moment it could be a problem for some users.

> Last, but not least, I intend to add kernel tests for any hid-bpf program we ship upstream. So that means that we will have regression tests for those devices, allowing for a better "no regression" rule.

That'd be a huge +1 to use HID-BPF, but the same could be done for UCLogic/DIGImend.

I acknowledge that I'm the one to blame for not adding  user-space tests for the devices I added support in the kernel, this is something I need to fix. At least I added KUnit tests... It's something.

I don't oppose to the idea of using HID-BPF or a simpler driver like Peter mentioned, but I don't see a clear advantage over growing an already existing solution.

>> It is the circular area inside the dial in devices like the [Deco Pro SW](https://www.storexppen.es/buy/deco-pro-sw.html). (Which by the way stopped working in working with "recent" libinput).
>>
> oh? that'd be a bug - isn't that a REL_WHEEL? ow what does it send?

@whot that circular area is a dial (REL_WHEEL) and a "touchpad" that sends mouse events. I'd try to open a bug report upstream after investigating the issue.

By the way, the XP-Pen Deco Pro S/M/MW and the Parblo A610 Pro report dial events throughout the pad event node.
These are the latest devices added to the kernel in the last ~2 years. I added them and own them, so let me know if you need my help testing any of your dial-related PRs.

whot:
> By the way, the XP-Pen Deco Pro S/M/MW and the Parblo A610 Pro report dial events throughout the pad event node.
These are the latest devices added to the kernel in the last ~2 years. I added them and own them, so let me know if you need my help testing any of your dial-related PRs.

If you can attach a `libinput record` for these devices to a libinput issue we can add them as test devices which makes things simpler to test in the future.

JoseExposito:
A quick summary as reference for the future:

- DIGImend will keep including fixes for devices.
- Fixes will be backported to the upstream kernel.
- About devices not present in DIGImend: HID-BPF offers so many advantages for developers and end users, that we'd start using it as the preferred way to support new devices.
- New devices might be added to DIGImend, but they won't be backported to the upstream kernel.

Closing this PR as this change would cause regressions.

PR #695, opened 2024-05-26
DIGImend: Remove udev tablet bit from XP-PEN mouse interface

JoseExposito:
XP-Pen devices expose 3 different event nodes to user-space:

 - One for the pen
 - A second one for the pad buttons and dial
 - A final one for their mouse

The "mouse" is a circular area inside the dial that looks like a circular touchpad but sends mouse events.

Both ID_INPUT_TABLET and ID_INPUT_MOUSE udev properties are set in this event node but only ID_INPUT_MOUSE should be set.

Add udev rules to remove ID_INPUT_TABLET in the devices with a "mouse".

@whot let me attach a recording of a XP-Pen dial so you have another device type for testing: [xp-pen-deco-pro-sw-dial.txt](https://github.com/DIGImend/digimend-kernel-drivers/files/15448096/xp-pen-deco-pro-sw-dial.txt).

Also, since you are here, do you know if it is possible to upstream these udev rules? These devices are in the kernel and used to work without these rules but at some point their "mouse" stopped working.


whot:
> Both ID_INPUT_TABLET and ID_INPUT_MOUSE udev properties are set in this event node but only ID_INPUT_MOUSE should be set.

I wonder if that's an effect of the libwacom rules going rogue. That vid/pid is in libwacom and libwacom produces this hwdb entry:
```
 # UGTABLET Deco Pro SW
  libwacom:name:*:input:b0003v28BDp0933*
   ID_INPUT=1
   ID_INPUT_TABLET=1
   ID_INPUT_JOYSTICK=0

  libwacom:name:* Pad:input:b0003v28BDp0933*
   ID_INPUT_TABLET_PAD=1

  libwacom:name:* Keyboard:input:b0003v28BDp0933*
   ID_INPUT_TABLET=0
```
Maybe libwacom needs a `Mouse` entry that removes `ID_INPUT_TABLET` again?

There's also this https://github.com/systemd/systemd/commit/0855ce67726f87a5a67b4fb536d58e0e4428a248 which labels this device as `ID_INPUT_TABLET_PAD` because it has `BTN_0`.

I suspect one of these two (probably both) are the cause for this. How does the Mouse event node look like in `libinput record`? I don't get that one with the recording here (neither with  hid-replay from the `hid:[]` in the recording.


> Also, since you are here, do you know if it is possible to upstream these udev rules?

Theoretically, you shouldn't need to. udev-builtin-input_id has its detection and that should work as long as you provide the right bits in a tablet. And those you need anyway to get libinput to work with the tablet.

libwacom has them partially for historical reasons and for the `ID_INPUT_TABLET_PAD` which didn't really exist in systemd until later. Once libwacom has the tablet added you're doubly assigned. but either way, I'm struggling to see why these days you'd need the udev rules at all here - various bugs nonwithstanding.

JoseExposito:
For reference, this is the recording of the mouse interface: 
[xp-pen-deco-pro-sw-mouse.txt](https://github.com/user-attachments/files/15748631/xp-pen-deco-pro-sw-mouse.txt)

It doesn't have BTN_* nor a wheel, so I think that your fix in libwacom should be enough to get it working again, thanks!

Closing this PR.

whot:
yep, that should do the trick and definitely a case of the libwacom hwdb being a bit too globby. Thanks

PR #698, opened 2024-06-25
Add support for XP-Pen Artist 16 2nd Gen

Krzysztof-WW:
Hi, I fixed hid descriptor for this tablet and everything works, pressure, buttons on stylus, tilt. Only buttons on tablet are reported as normal keyboard but that can be remapped later in the system.


JoseExposito:
Hi @Krzysztof-WW 

I'd recommend to create a HID-BPF driver instead of adding it here. DIGImend is almost unmaintained and HID-BPF is the easiest option to get your changes in the upstream kernel.

Here is a quick summary of the decission:
https://github.com/DIGImend/digimend-kernel-drivers/pull/693#issuecomment-2126861337

And here is an example adding support for another XP-PEN device:
https://gitlab.freedesktop.org/libevdev/udev-hid-bpf/-/merge_requests/88

And a tutorial:
https://libevdev.pages.freedesktop.org/udev-hid-bpf/tutorial.html

You did most of the work already, it'd be as easy as copy and paste the linked code and change the descriptors. Let me know if you need help with it.

Krzysztof-WW:
I think I should be able to write a HID-BPF driver. It really looks like I only need to change descriptors and buttons in example you provided to have a working driver.
Also thank you for the info on HID-BPF, because I didn't know there was better a alternative for DIGImend.

PR #707, opened 2024-11-25
Update to Linux 6.12

Stroopwafe1:
This patch allows the digimend kernel drivers to build on Linux kernel 6.12.
The reason it failed before is because of a header rename from `<asm/unaligned.h>` to `<linux/unaligned.h>`. And the struct of the hid_driver changed requiring a const pointer for the `report_fixup`

### DIFF
--- a/hid-kye.c
+++ b/hid-kye.c
@@ -579,7 +579,7 @@ static __u8 *kye_consumer_control_fixup(struct hid_device *hdev, __u8 *rdesc,
 	return rdesc;
 }
 
-static __u8 *kye_report_fixup(struct hid_device *hdev, __u8 *rdesc,
+static const __u8 *kye_report_fixup(struct hid_device *hdev, __u8 *rdesc,
 		unsigned int *rsize)
 {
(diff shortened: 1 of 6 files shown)


kenhys:
It would be nice if it also still supports 6.11 (`#if LINUX_VERSION_CODE < KERNEL_VERSION(6, 12, 0)`)

Stroopwafe1:
I have updated the code with a check for the linux kernel, and I tested this in a VM with kernel version `6.9.3-76060903-generic` and it built fine, as well as on my regular hardware using kernel `6.12.4-arch1-1`. Let me know if there's anything else.

SohamG:
Hey! I just updated to 6.12 and had to pull this driver from my system. I was wondering if this PR would land soon (given the primary maintainers resignation blog post)? I was considering adding this as a backport to my linux distro in the meanwhile (NixOS)

JoseExposito (approved):
I was fixing the same issue, but I realized you fixed it already. Thanks!

PR #714, opened 2025-06-14
fix: Rename `del_timer_sync()` to `timer_delete_sync()`

tokenicrat:
The timer function `del_timer_sync()` was renamed to `timer_delete_sync()` in November 2022 (<https://lkml.org/lkml/2022/11/23/1482>), and a wrapper for it was recently removed. This prevents DIGImend from compiling on Linux 6.15.0 and later.

The issue can be solved by just replacing the name. Validated on Arch Linux (Linux 6.15.2).

### DIFF
--- a/hid-uclogic-core.c
+++ b/hid-uclogic-core.c
@@ -491,7 +491,7 @@ static void uclogic_remove(struct hid_device *hdev)
 {
 	struct uclogic_drvdata *drvdata = hid_get_drvdata(hdev);
 
-	del_timer_sync(&drvdata->inrange_timer);
+	timer_delete_sync(&drvdata->inrange_timer);
 	hid_hw_stop(hdev);
 	kfree(drvdata->desc_ptr);
 	uclogic_params_cleanup(&drvdata->params);
--- a/hid-uclogic-core.c
+++ b/hid-uclogic-core.c
@@ -25,6 +25,14 @@
 #include "compat.h"
 #include <linux/version.h>
 
+/* 
+ * Linux kernel >= 6.1.84 no longer includes wrapper for deprecated 
+ * del_timer_sync
+ */
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 1, 84)
+#define del_timer_sync timer_delete_sync
+#endif
+
 /**
  * uclogic_inrange_timeout - handle pen in-range state timeout.
  * Emulate input events normally generated when pen goes out of range for
@@ -491,7 +499,7 @@ static void uclogic_remove(struct hid_device *hdev)
 {
 	struct uclogic_drvdata *drvdata = hid_get_drvdata(hdev);
 
-	timer_delete_sync(&drvdata->inrange_timer);
+	del_timer_sync(&drvdata->inrange_timer);
 	hid_hw_stop(hdev);
 	kfree(drvdata->desc_ptr);
 	uclogic_params_cleanup(&drvdata->params);


Eason0729:
I thinks a better approach would be using `#if LINUX_VERSION_CODE >= KERNEL_VERSION(6,15,0)`, which enable compilation before the rename

```c
#if LINUX_VERSION_CODE >= KERNEL_VERSION(6,15,0)
#define del_timer_sync timer_delete_sync
#endif
```

tokenicrat:
> I thinks a better approach would be using `#if LINUX_VERSION_CODE >= KERNEL_VERSION(6,15,0)`, which enable compilation before the rename
> 
> ```c
> #if LINUX_VERSION_CODE >= KERNEL_VERSION(6,15,0)
> #define del_timer_sync timer_delete_sync
> #endif
> ```

Definitely it's safer, but as all mainstream desktop Linux distributions (including Ubuntu, Fedora, Debian, Arch Linux, etc.) under maintenance are shipped with Linux 6.x.y, I think it's okay to use the new name only.

kenhys:
Accurately, it should be because timer_delete_sync was introduced since 6.1.84:

```c
#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 1, 84)
#define del_timer_sync timer_delete_sync
#endif
```

By the way, it seems that DIGImend kernel drivers try to support older kernel (I don't know whether that is actually still buildable or not) - see such as README.md , hid-uclogic-core.c, hid-uclogic-params.c.
Shouldn't we respect it?

ruffianlabs:
as [tokenicrat](https://github.com/tokenicrat) said, I tried & replaced the file  `/usr/src/digimend-13/hid-uclogic-core.c` and, made the renaming from : 
`del_timer_sync(&drvdata->inrange_timer);` TO `timer_delete_sync(&drvdata->inrange_timer);`

and, updated the linux & linux-headers. and, it worked. No errors.

My setup : Endeavour OS (Linux 6.16.8-arch3-1 & nvidia/580.82.09 | DE: KDE Plasma 6.4.5 | WM: KWin (Wayland))

Thank you [tokenicrat](https://github.com/tokenicrat) & everyone who worked on this.

AlexandraMercier:
I tried the fix using @kenhys code, and I confirm it builds on ArchLinux (Linux 6.17.7.arch1-1).
I second his opinion that we should keep retro-compatibility with older systems.

arkadesOrg:
**just built and tested with HUION H950P on** 
Kernel Version: 6.18.6-arch1-1 (64-bit)
KDE Plasma Version: 6.5.5
KDE Frameworks Version: 6.22.0
Qt Version: 6.10.1
Graphics Platform: Wayland

PS: works much better than manufacturer driver btw ❤️

tokenicrat:
Now it's using @kenhys version for backward compatibility.

It's been 8 months, and the repository seems not to be actively maintained. If possible I recommended everyone here to use more recent solutions like OpenTabletDriver.

arkadesOrg:
I am pretty sure I will never install all the OpenTabletDriver .NET bloat just to use a simple tablet.

The latest patch is now delivered with Archlinux User Repo https://aur.archlinux.org/packages/digimend-kernel-drivers-dkms.

The Archlinux User Repo git-version https://aur.archlinux.org/packages/digimend-kernel-drivers-dkms-git does now temporarly point to the downstream branch [arkadesOrg:hotfix](https://github.com/arkadesOrg/digimend-kernel-drivers/tree/hotfix)

